PR #4, opened 2026-07-03
fix(macos): stop false "uncaptured credentials" prompt on every switch

xingfanxia:
## Problem

On macOS, the CLI reconcile prompt fires on **every** `clauth <name>` switch:

> active profile 'X' has uncaptured credentials in ~/.claude (a re-login or token rotation). capture them into 'X' and switch to 'Y'? [Y/n]

…even when nothing has actually changed.

## Root cause

`classify_link_at` treats **any** non-symlink at `~/.claude/.credentials.json` as `Diverged` without inspecting its content:

```rust
if !meta.file_type().is_symlink() {
    return Ok(LinkState::Diverged);
}
```

On macOS, Claude Code reads its login from the **Keychain** and mirrors a copy into `~/.claude/.credentials.json` as a plain 0600 file, rewriting it on every run. That replaces clauth's symlink with a regular file whose **content is identical** to the active profile's stored credentials. So `classify_link_at` reports `Diverged` on every switch, and `switch_profile_cli` prompts — a pure false positive (there is nothing new to capture).

I verified on my machine that the live `~/.claude/.credentials.json` and the active profile's `credentials.json` were byte-identical (same access token) while the prompt kept firing.

## Fix

For a non-symlink, compare the live file's OAuth **access token** to the profile's stored token instead of trusting symlink identity:

- token matches → `LinkedTo` (it's Claude Code's mirror — nothing to capture, no prompt)
- token differs → `Diverged` (a genuine re-login / rotation worth capturing — prompt as before)

This is cross-platform and strictly more correct than symlink identity (on Linux, Claude Code writes `.credentials.json` directly too, so content is the right signal there as well). Genuine re-login detection is preserved — the existing `relogin_is_diverged_and_not_first_login` / overwrite-branch tests still pass.

## Tests

Two regression tests added:
- plain file whose token **matches** the profile → `LinkedTo` (the bug)
- plain file whose token **differs** → `Diverged` (real re-login still detected)

`cargo test` green (451 passed), `cargo clippy --all-targets -- -D warnings` clean.

---

Context: found while building a macOS Keychain-backed fork (the divergence prompt made every switch noisy on macOS). Happy to adjust wording/approach if you'd prefer a different signal (e.g. also comparing the refresh token). 🙏

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -56,8 +56,30 @@ pub(crate) fn switch_profile(config: &mut AppConfig, name: &str) -> Result<()> {
         if config.is_active(name) {
             return Ok(());
         }
+        // Is the outgoing live file an UNCAPTURED CC re-login? `snapshot_active_
+        // credentials` deliberately skips capturing that case (Diverged & not a
+        // first-login), so dropping it would strand a fresh `/login` chain — keep
+        // the non-force refuse-guard there. Every other state is captured or
+        // adoptable by the snapshot below, so force the relink: on macOS the live
+        // `.credentials.json` is always a regular-file Keychain mirror of the
+        // active account and thus legitimately differs from the target, which the
+        // non-force guard's live-vs-target byte check would wrongly reject on every
+        // switch. (Interactive callers already route a real divergence to the
+        // reconcile path, so this branch is only reachable uncaptured via the
+        // scheduler — where refusing, not dropping, is the safe outcome.)
+        let uncaptured_relogin = match config.state.active_profile.as_deref() {
+            Some(active) => {
+                matches!(classify_credentials_link(active)?, LinkState::Diverged)
+                    && !is_first_login(active)?
+            }
+            None => false,
+        };
         snapshot_active_credentials(config)?;
-        link_profile_credentials(name)?;
+        if uncaptured_relogin {
+            link_profile_credentials(name)?;
+        } else {
+            force_link_profile_credentials(name)?;
+        }
         finish_switch(config, name)
     })
 }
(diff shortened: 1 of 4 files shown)


xingfanxia:
Added a second commit (dc46b16) that MUST land with the classify fix — otherwise this PR would introduce a regression on macOS:

Once `classify_link_at` stops flagging the identical-content mirror as `Diverged`, `switch_profile_cli` takes the plain `switch_profile` path instead of the reconcile path. `switch_profile` then hit `link_profile_credentials`'s refuse-guard (`live bytes != target bytes`) and bailed with **"refusing to replace .credentials.json — live file differs from profile 'X'"** on *every* `clauth <name>` — because on macOS the live file is always a regular-file Keychain mirror of the active account and legitimately differs from the switch target.

Fix: `switch_profile` now snapshots the outgoing creds (which captures every state except an uncaptured re-login) and then **force**-relinks — safe, since the live file was just captured into the active profile. The non-force refuse-guard is kept only for a genuine uncaptured re-login (`Diverged && !first_login`), reachable only via the scheduler (no divergence pre-check), where refusing rather than dropping the fresh `/login` chain is the safe outcome. Interactive callers already route real divergence to the reconcile path, so their behavior is unchanged. Regression test included.

uwuclxdy:
reviewed and ran it locally: clippy clean, 452 tests green, so this is going in.

traced the switch state machine. computing `uncaptured_relogin` before the snapshot is the right call. the else-branch `force_link` never drops a capturable state either: the macOS mirror reads as `LinkedTo`; a genuine uncaptured re-login stays behind the `link_profile_credentials` refuse-guard; a first-login gets adopted before the relink.

two notes, neither blocking:

1. the compare looks at the access token only, where `credentials_diverged` uses access plus refresh. that's fine in practice since the pair rotates together, so i folded in one guard on merge: an empty `Some("")` access token now classifies as `Diverged` rather than `LinkedTo`, matching the completed-login intent of `is_first_login`.

2. this quiets the false prompt only until the next token refresh. on macOS CC rotates into the Keychain with a single-use token, so once the stored chain and the live mirror drift apart, the `Diverged` path comes back. that residual is the Keychain-mirror problem underneath, which i'm tracking for the macOS milestone (post-login harvest plus a `security`-CLI write). no regression here, since before this every switch bailed.

Issue #1, opened 2026-07-03
MacOS Keychain Support


xingfanxia:
```
Why (confirmed from clauth's own source): clauth is file-only and has zero macOS Keychain support. It snapshots ~/.claude/.credentials.json, but on your Mac that file doesn't exist — your login is in the Keychain. clauth's read returns "nothing" instead of erroring, so it silently captured empty profiles.
``` 

MacOS by default uses this keychain to store credential

uwuclxdy:
hi @xingfanxia, thanks for reading the source before filing. your analysis is right: clauth only reads `~/.claude/.credentials.json`. when that file is missing it happily saved an empty profile and called it a success.

two things are in the works already:

**loud failure instead of silent empty profiles**: capture now refuses when it finds nothing to capture, with an error naming the keychain situation.

**`clauth login <profile>`**: a new command that creates the profile, then runs Claude Code's own `/login` inside that profile's private runtime so the completed login lands in the profile. no snapshotting of your current session at all. whether this fixes macOS outright depends on one detail I can't test without a Mac, which brings me to a favor.

could you run two quick checks on your machine?

1. `CLAUDE_CONFIG_DIR=/tmp/cc-probe claude`, sign in with `/login`, exit, then check whether `/tmp/cc-probe/.credentials.json` exists. if CC writes a real file when given a custom config dir, `clauth login` should already work on macOS as-is.
2. `security find-generic-password -s "Claude Code-credentials" -w` (please don't paste the output anywhere, it contains your live tokens!). just say whether it looks like the same JSON shape as a `.credentials.json` file, meaning a `claudeAiOauth` object with `accessToken`/`refreshToken`.

answer 1 decides whether a keychain backend is even needed; answer 2 shapes it if it is. lmk what you find, or if you'd rather see this solved a different way ^^

xingfanxia:
> hi [@xingfanxia](https://github.com/xingfanxia), thanks for reading the source before filing. your analysis is right: clauth only reads `~/.claude/.credentials.json`. when that file is missing it happily saved an empty profile and called it a success.
> 
> two things are in the works already:
> 
> **loud failure instead of silent empty profiles**: capture now refuses when it finds nothing to capture, with an error naming the keychain situation.
> 
[...]

PR #18, opened 2026-07-05
feat(macos): route account switches through the Keychain

xingfanxia:
## What

Route macOS account switches through the login **Keychain**, so a switch actually takes effect on a running Claude Code session.

## Why

On macOS, Claude Code reads its OAuth login from the login Keychain — a generic password at service `Claude Code-credentials`, account = the OS login name — **not** from `~/.claude/.credentials.json`. So clauth's symlink swap is cosmetic on macOS: Claude Code keeps reading the Keychain, and the switch silently doesn't apply to a running session. This makes the switch real by mirroring the target profile's stored login into the Keychain (and dropping it on clear).

Follow-on to the merged macOS divergence fix (#4).

## How

- **`src/keychain.rs`** (new) — write / delete Claude Code's login item via Apple's `/usr/bin/security` CLI.
  - **Why the CLI, not the `security-framework` crate:** macOS binds a Keychain *"Always Allow"* grant to the **calling binary's code signature**. A `cargo build` / `cargo install` binary carries an ad-hoc signature that changes on every rebuild, so a grant against clauth never persists — every switch would re-prompt. Routing the write through Apple's stable, code-signed `/usr/bin/security` binds the one-time grant to *that* binary, so it sticks across clauth rebuilds — no code-signing setup required. (Same approach CCSwitcher uses.)
  - A **20 s wall-clock deadline** kills (and reaps) a stuck `security` invocation — an unanswered "Always Allow" ACL prompt or a locked keychain — so it can't wedge the caller while it holds the state lock.
- **`src/claude.rs`** — wires the write on `link_profile_credentials` / `force_link_profile_credentials`, and a delete on `clear_claude_credentials`, each behind `#[cfg(target_os = "macos")]` + `keychain::enabled()`.

## Design notes

- **Write-only.** clauth stores every profile's credentials as a file (`~/.clauth/profiles/<name>/credentials.json`) and never *reads* Claude Code's Keychain item in any automatic path — reading it raises a macOS access prompt on every call. The Keychain is touched only to make a switch real (write on link, delete on clear).
- **No-op when a profile has no stored `credentials.json`** (e.g. a `base_url` profile, or an OAuth profile not yet logged in) — the existing Keychain login is left untouched.
- **macOS-only, cfg-gated.** The `mod keychain;` declaration and all call sites are `#[cfg(target_os = "macos")]`, so non-macOS builds are byte-for-byte unchanged.

## Testing

- `keychain::enabled()` returns `false` under `#[cfg(test)]`, so the whole suite keeps the file/symlink model and **never touches the operator's real `Claude Code-credentials` item**.
- The `KC-1` round-trip test (`tests/inline/keychain.rs`) exercises the real `security` read/write/delete path against a **throwaway, process-unique service name**, and is `#[ignore]`d (it still mutates the login Keychain as a side effect):
  ```
  cargo test keychain_round_trip -- --ignored
  ```
- `cargo build`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` are all green (458 passed, 2 ignored). Diff is **+374 / −0**, fully additive.

## Companion project

I've been running this against a macOS menu-bar app that reads the daemon's status feed and drives switches — happy to point at it if a daemon/`status.json` contribution follows.

### DIFF
--- a/src/claude.rs
+++ b/src/claude.rs
@@ -116,6 +116,21 @@ pub(crate) fn read_claude_credentials() -> Result<Option<ClaudeCredentials>> {
     read_json_file(&path).map(Some)
 }
 
+/// macOS: mirror a profile's stored OAuth login into the Keychain so Claude Code
+/// (which reads the Keychain, not the file) actually switches account. No-op when
+/// the profile has no stored `credentials.json` (a base_url profile, whose
+/// endpoint+token come from `settings.json`, or an OAuth profile not yet logged
+/// in) — the existing Keychain login is left untouched in that case.
+#[cfg(target_os = "macos")]
+fn keychain_write_profile(name: &str) -> Result<()> {
+    let path = profile_dir(name)?.join("credentials.json");
+    if !path.exists() {
+        return Ok(());
+    }
+    let creds: ClaudeCredentials = read_json_file(&path)?;
+    crate::keychain::keychain_write(&creds)
+}
+
 #[cfg(unix)]
 pub(crate) fn create_symlink(target: &Path, link: &Path) -> Result<()> {
     std::os::unix::fs::symlink(target, link).context("Failed to create credential symlink")
@@ -164,6 +179,11 @@ pub(crate) fn link_profile_credentials(name: &str) -> Result<()> {
                 std::fs::create_dir_all(parent)?;
             }
             create_symlink(&target, &link)?;
+            // macOS: make the switch real — Claude Code reads the Keychain.
+            #[cfg(target_os = "macos")]
+            if crate::keychain::enabled() {
+                keychain_write_profile(name)?;
+            }
         }
 
         Ok(())
@@ -175,6 +195,12 @@ pub(crate) fn clear_claude_credentials() -> Result<()> {
         let link = claude_credentials_path()?;
         if link.symlink_metadata().is_ok() {
             std::fs::remove_file(&link).context("Failed to remove .credentials.json")?;
+        }
+        // macOS: also drop the live Keychain login so Claude Code can't spend the
+        // account (parity with removing the credential file).
+        #[cfg(target_os = "macos")]
+        if crate::keychain::enabled() {
+            crate::keychain::keychain_delete()?;
         }
         Ok(())
     })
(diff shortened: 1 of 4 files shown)


uwuclxdy:
Merged, thanks! Reviewed statically plus what Linux allows: un-gated `mod keychain` for a one-off `cargo check` (compiles clean) and ran both deadline tests (pass, child killed + reaped near the deadline). The argv-tradeoff doc, the `$USER` vs `"unknown"` account pinning, the test-suite isolation: all hold up.

Two notes from review, neither blocking:

- `security -i` (subcommand + `-w` value on stdin) would keep the token out of exec logs (EDR `es_event_exec_t`), which the current doc does not weigh. I suspect its word-tokenizer mangles the JSON's embedded `"` without extra quoting; probably why argv won. If you ever check, a sentence in the `write_at` doc would settle it.
- `clear_claude_credentials` now deletes whatever the keychain item holds at that moment, including a chain CC rotated after the last capture, or a login clauth never wrote. Inherent to write-only (reading prompts), so I documented the window in a comment instead of changing behavior (`b9d932e`).

The rotation-coherence residual stays open on #1: CC refreshing into the keychain mid-session can still drift from the profile's file chain. If you get a chance to observe what a CC-side refresh does to the keychain item vs the mirror file, that decides whether there is anything left to build.

xingfanxia:
Checked the `security -i` question on-device (macOS 15, throwaway service names only). Your suspicion is half right — and the half that's wrong makes it usable:

- an **unquoted** `-w` value splits on whitespace, so any payload with spaces is a usage error (compact serde JSON survives only by luck)
- but `-i` has a real escape convention: wrap the value in `"…"` with `\` → `\\` and `"` → `\"`, and arbitrary JSON — spaces, embedded quotes, backslashes — round-trips **byte-identical**
- `-U` behaves identically through `-i`, and the inner command's exit code propagates as `security`'s own (`0` / `44` errSecItemNotFound / `2` usage), so error handling needs no changes
- one caveat: an embedded `\n` in a value parses as a second command (line protocol), so it has to be refused up front

Since that closes the `es_event_exec_t` residual cleanly, I went one step past the doc sentence — #21 switches `write_at` to `-i` + stdin with the tokenizer rules documented on the fn, a `security_quote` helper (newline refusal included), and the round-trip test re-run against a real Keychain with a hostile spaces+quotes+backslash secret. `delete` stays argv (secret-free). Grant binding unchanged — the caller is still `/usr/bin/security`. ^^

Issue #3, opened 2026-07-03
Instead of saving from current authed session, open a new oauth sign in to auth


xingfanxia:
Instead of saving from current authed session, open a new oauth sign in to auth

uwuclxdy:
hi @xingfanxia, this is implemented as `clauth login <profile>`: it creates the profile and runs Claude Code's own `/login` inside that profile's runtime, so the login lands in the profile without touching your current session. it ships with the next release. the macOS caveat from #1 applies here too, the two checks I asked for over there would confirm it works on Mac as-is. lmk if that shape works for you ^^

Issue #7, opened 2026-07-04
Enhancement: Capturing a new login, can't overwrite an existing account/profile name


edgan:
I created `claude@domain.com`, but then logged into that account while on `user@domain.com`. I couldn't pick `claude@domain.com`. I tried entering `claude@domain.com` and it told me it already exists. Both the ability to pick and the ability to enter `claude@domain.com`. Along with a dialog to allow overwrite would be nice.

Edit: Version `master`, `71ce018d7ad516f61dd0765181141a5018b7a605`.

uwuclxdy:
Plan for fixing this: Typing an existing name will open an overwrite confirm dialog instead of rejecting with a toast notification. Overwrite will swap the credentials in place while keeping the usage history, env, model settings and chain position. The picker might come later since this covers the main thing. LMK if that's okay.

PR #19, opened 2026-07-05
feat(oauth): self-contained browser OAuth login

xingfanxia:
## What

Replace `clauth login` with a **self-contained browser OAuth sign-in** that reproduces Claude Code's own PKCE flow, so it works on macOS (and identically on Linux/Windows).

## Why

Today `clauth login <profile>` runs Claude Code's own `/login` inside the profile's runtime and adopts the credentials file the runtime watchdog reads. **On macOS that silently fails:** a fresh `/login` under a custom `CLAUDE_CONFIG_DIR` writes only a per-config-dir *hashed* Keychain item (`Claude Code-credentials-<hash>`), never `.credentials.json` — so clauth reports *"no login detected — profile still has no credentials"* and the profile stays empty. (Verified on-device; see #1, #3.)

## How

A loopback OAuth flow that mints the tokens itself, with no dependency on Claude Code's platform-specific credential storage:

1. Generate a PKCE `code_verifier` + `state`, open the system browser to `claude.com`'s authorize page.
2. Bind an RFC 8252 loopback listener on `127.0.0.1:<random port>` to catch the redirect.
3. Exchange the returned code at `platform.claude.com` for a fresh token pair.
4. Verify the token against `/profile` (a `401` means a dud login) and stamp the real plan tier.
5. Write the tokens straight into a fresh profile.

Works identically on Linux, macOS, and Windows.

## Changes

- **`src/oauth_login.rs`** (new) — the PKCE flow + loopback listener (+ inline tests).
- **`src/oauth.rs`** — `exchange_code` + the login token endpoint; `CLIENT_ID` is now `pub(crate)`.
- **`src/usage/fetch.rs`** — `probe_subscription_type`: reuses the usage request path to confirm the minted token works and derive the plan tier, so a new profile shows its real plan immediately.
- **`src/platform.rs`** — `open_url` (default-browser launch, detached stdio).
- **`src/main.rs`** — `cmd_login` rewired to the browser flow; help text updated.
- Drops the now-obsolete `actions::ensure_login_profile` and its tests (login no longer pre-creates a blank profile).
- **docs** — README feature line + SECURITY host table (`claude.com/…/authorize`, `platform.claude.com/…/token`) and a "what acts on your behalf" entry.

## Dependency

Pins `getrandom = "0.4"` as a **direct** dep for the PKCE `code_verifier` + `state` CSPRNG. `0.4.x` is already in the tree transitively, so this adds no new resolved version; no `unsafe` in-crate.

## Testing

- `cargo build`, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` all green (**460 passed, 1 ignored**).
- `tests/inline/oauth_login.rs` covers the PKCE challenge/verifier, the auth-URL construction, and the loopback redirect parsing.
- No network is touched in tests — the flow's IO boundaries are exercised with local fixtures.


uwuclxdy:
Merged, thanks! The PKCE core reviewed clean: real refresh token enforced by the parse (`refresh_token` is a required `String`), `state` generated + verified, `127.0.0.1`-only bind, RFC 7636 vector test. Landed four small commits on top:

- `f6376e0`: windows `open_url` was broken. `cmd /C start "" <url>` reaches cmd unquoted (std only quotes an arg on space/tab/quote), so cmd splits the URL at the first `&` and runs the tail as commands. Reproduced in rust on a real windows box. The fix goes through `rundll32 url.dll,FileProtocolHandler` instead of quoting: no cmd tokenizer at all, no `%xx` variable-expansion exposure. Compiles on the box; the full browser round-trip still needs an interactive session there, so that leg is pending.
- `830af4f`: socket-pair tests for the callback security paths (`state` mismatch abort, `error` param abort, favicon + missing-code ignored, deadline timeout).
- `9dbf636` + `b9d932e`: `getrandom` patch pin, dropped the `.agent/PROGRESS.md` ref that did not ship.

The one open question: whether a pair minted at `platform.claude.com` refreshes at `api.anthropic.com`. Rotation depends on it, and I cannot mint a pair from here without burning a fresh login. If you can run `clauth login <tmp>` on current `mommy` and force-rotate once (`t` on the profile), that closes it.

Btw `clauth login <existing>` now bails "already exists" (the reuse path went away with the rewire). Fine imo, it never worked on macOS anyway; routing existing names through the #7 overwrite path is parked as a follow-up.

PR #24, opened 2026-07-07
fix(auth): rotation coherence — clauth owns the active Keychain chain

xingfanxia:
Closes the rotation-coherence residual you kept #1 open for — with the on-device observation you asked for, because it fired on my machine yesterday, in **both** directions.

## The incident (macOS, 2026-07-07)

The running `claude` re-reads the Keychain per request, so the live login and clauth's profile copy are two independent refresh chains of ONE OAuth family — and the refresh token is single-use: whoever refreshes first revokes the other side.

- **clauth wins the race** (idle session): the poller's 401-driven rotation revoked the Keychain copy's refresh token. The running `claude` died at its stale token's expiry — `Not logged in · Please run /login` — while every clauth copy stayed green. Nothing in clauth could even *see* it: the write-only Keychain integration monitors the profile copy, which was healthy.
- **CC wins the race** (active session): CC refreshes lazily near expiry and revokes clauth's stored pair; the poller's next refresh comes back `invalid_grant` on a perfectly healthy account.

## The fix — clauth deliberately wins, and keeps the live login fresh

1. **Mirror on rotate**: `apply_rotated_tokens_locked` (the choke point every rotation site already shares) now mirrors the ACTIVE profile's fresh pair into the Keychain, inside the same locked section as the persist so a concurrent switch can't re-point the item in between. Failure is loud but non-fatal (the rotation itself is durable; the next rotation or switch retries).
2. **Proactive lead**: `fetch_with_rotation` rotates the active Keychain-installed profile ~30 min *ahead* of expiry (`ACTIVE_ROTATE_LEAD_MS`) instead of waiting for a 401. Reactive rotation always loses the race to a live session; with the lead + mirror, CC never sees an expiring token, never refreshes, and there is only one chain. A bigger lead costs an *earlier* rotation, not an extra one — each refresh mints a full-lifetime token.

Once clauth owns the chain, the divergence class you and #4 dealt with shrinks too: CC stops rotating into the Keychain mid-session, because it never has a reason to.

## The subtle bit — the foreign-login gate

CC rewrites `~/.claude/.credentials.json` as a regular-file mirror of the Keychain, so the moment a rotation lands, `classify_credentials_link` reports `Diverged` against the NEW stored token even though the live login is just our own chain one step behind. A naive divergence gate would skip the mirror exactly when it matters. `live_login_is_foreign` tells the two apart: a live token matching the *pre-rotation* pair is our own stale mirror (mirror proceeds); a token matching neither pair is a real CC re-login (never overwritten); unclassifiable state is treated as foreign.

## Scope / safety

- No-op on Linux/Windows (`keychain_live()` is false — the symlinked profile file IS the live credential, no second chain to race) and for inactive profiles and under `cfg(test)`.
- Early rotation bails (guard busy, live `clauth start` session, missing refresh token, transient refresh failure) fall back to the **plain fetch** in proactive mode — proactivity never costs a live usage poll.
- Unknown expiry never rotates proactively (same conservatism as the existing gates).
- 596 tests, clippy + rustfmt clean. New unit tests for the lead-window predicate and all four `live_login_is_foreign` cases (temp-home sandbox; no real Keychain or credentials touched).

I can't exercise the Linux side beyond the suite, mirroring your constraint in reverse — but the non-macOS diff is inert by construction.

Heads-up: this touches `fetch_with_rotation`, same region the AUTH-1 dead-login PR (#1 offer) will touch — happy to rebase whichever lands second.


uwuclxdy:
thanks for your efforts to make clauth work on macos and the prs! can i ask you to double check if the proactive 30 min before expiry token refresh is the only way to solve this? i dont like it too much since it depends on claude code not refreshing sooner for any reason (they can change cc behavior and the issue will be back). please try to find a better, more future proof and stable solution. thanks again

xingfanxia:
Fair challenge — the 30-min lead was a race heuristic, and any fixed margin silently loses to a future CC that refreshes earlier. Pushed a v2 (93b143e) that makes correctness independent of timing. Three layers, and only the last is timing-related at all:

**1. Adopt, don't fight (the future-proof core).** The insight I'd missed: on macOS Claude Code maintains `~/.claude/.credentials.json` as a regular-file mirror of its Keychain login — a *prompt-free* read path to CC's current pair, no Keychain read needed (which stays write-only). So when CC refreshes first — schedule change, clauth downtime, lost race, anything — the poller now **adopts** CC's fresher pair from that mirror instead of spending our single-use refresh token against a family CC just superseded. `try_adopt_live_rotation` runs at rotation-leg entry and again on a refresh failure, every poll, so a lagging store self-heals as soon as the mirror catches up (at latest on CC's next launch). Whatever refresh behavior a future CC ships, the answer is the same: concede and adopt.

**Identity guard**: an unattended adopt must never capture a *different* account (a manual CC `/login`) into the profile. The mirror token's account uuid (`/api/oauth/profile`) must match the profile's cached anchor — seeded at `clauth login`, backfilled on every adopt — or the stored token's own uuid while it still authenticates. Unprovable identity refuses the adopt and leaves it to the existing TUI divergence flow. Injected identity fn, so all six gate branches are unit-tested offline.

**2. Mirror-on-rotate** (unchanged from v1): when *we* rotate the active profile, the fresh pair lands in the Keychain in the same locked section. CC re-reads the Keychain per request, so our rotations are invisible to it.

**3. The lead, reframed and demoted.** Replaced the magic 30 minutes with an interval-derived value (3 poll intervals, 3-min floor) maintaining one invariant: *the Keychain never holds an expired token while the poller is alive*. Under CC's current lazy model (refresh only when the token it just read looks spent) that removes CC's reason to refresh entirely — but nothing *depends* on that anymore: if CC refreshes anyway, layer 1 absorbs it. Happy to drop layer 3 entirely if you'd rather have zero ahead-of-expiry refreshes; it's now an optimization (fewer adopt events), not a correctness mechanism.

603 tests (7 new), clippy + rustfmt clean. Note for sequencing with #25: on the fork where both live together, a successful adopt also clears the `auth_broken` quarantine (the account was never dead — we just lost a race); that line is omitted here since this branch doesn't have the quarantine, and whichever PR lands second picks it up in the rebase.

xingfanxia:
Pushed one more commit (`48179cb`): the identity anchor now **backfills on the hourly `/profile` tier fetch** (write-if-missing, zero extra HTTP — the response is already parsed for the plan tier).

Why: we hit the gap live yesterday. A profile that predates login-time anchor seeding has no `account_id.json`; once its stored pair fully died, `try_adopt_live_rotation` could not prove the diverged live login was the same account and the profile quarantined instead of healing — even though Claude Code held a perfectly healthy fresher pair the whole time. With the backfill, any successful fetch while the stored token is still alive persists the anchor, so adoption stays provable after the chain dies. Blank uuid refused, existing anchors never overwritten (login remains the authoritative re-seeder). 606 tests green.

uwuclxdy:
Requesting changes. The adopt-don't-race core is correct and can stay as is; the requests below are about packaging and hardening.

- Make layer 3 an opt-in config toggle, off by default: `preemptive_rotation` (`#[serde(default, skip_serializing_if = "is_false")]`, matching the existing toggles), surfaced as "Preemptive rotation" in the Config tab and gating `proactive_rotation_due`. Stock clauth should stay strictly lazy. Adoption together with mirror-on-rotate already provides the correctness; the early refresh is an optimization some setups may want.
- `README.md:73` still describes the v1 design ("~30 min ahead", "clauth deliberately wins the race"). The actual lead is `interval*3` with a 3 min floor, about 4.5 min at the default 90s cadence. The race framing was dropped in v2. Adoption is not mentioned at all. Please reframe the bullet around adoption, describing the early lead as the opt-in toggle.
- `apply_rotated_tokens_locked` holds the cross-process state flock across the `/usr/bin/security` shell-out (up to its 20s kill deadline). A hung `security` child would block every other clauth instance's state writes for that window. Please capture `creds` under the lock and write the Keychain after releasing it.
- `try_adopt_live_rotation` mutates stored credentials outside the `RotationGuard`. Please move the adopt under the guard, or document why the state flock alone is sufficient.
- Please un-gate `mod adopt_live_rotation` from `cfg(target_os = "macos")` (keeping `keychain_mirror_gate` gated). The adopt checks are platform-independent and should run on the Linux leg as well.
- `scheduler.rs:321` doc-links `[ACTIVE_ROTATE_LEAD_MS]`, which the v2 rename removed. Please point it at `active_rotate_lead_ms`.

Let me know if the toggle scope seems wrong; otherwise this is close to mergeable.

README.md
```diff
@@ -70,7 +70,7 @@ Most account tools do one half. clauth pairs instant **switching between multipl
 
 ### Automate & stay safe
 
-- **Automatic token refresh**: OAuth refresh tokens are single-use, so rotation stays lazy. A stale access token rotates the moment a usage query 401s, never ahead of time. <kbd>t</kbd> force-rotates every account.
+- **Automatic token refresh**: OAuth refresh tokens are single-use, so rotation stays lazy for every account except one. A stale access token rotates the moment a usage query 401s; the **active** account on macOS instead rotates ~30 min *ahead* of expiry and mirrors the fresh pair straight into the Keychain — the running `claude` shares that single-use chain, and whoever refreshes first revokes the other side, so clauth deliberately wins the race and keeps the live login permanently fresh (rotation coherence, #1). <kbd>t</kbd> force-rotates every account.
```
uwuclxdy: This bullet still describes the v1 design. The lead is not ~30 min at the default cadence: `active_rotate_lead_ms` is `interval*3` with a 3 min floor, about 4.5 min at the default 90s interval (30 min would require a 10 min polling interval). "clauth deliberately wins the race" is the framing v2 dropped; the adopt path that actually provides the correctness is not mentioned. Please reword this around adoption (clauth concedes and adopts CC's fresher pair, identity-guarded), describing the early lead as the `preemptive_rotation` opt-in.

src/usage/scheduler.rs
```diff
@@ -264,6 +316,10 @@ fn load_cached_with_status(name: &str, status: FetchStatus) -> (Option<UsageInfo
 /// Returns a [`FetchOutcome`]: the rotated pair for the caller's `TokenList`
 /// sync, the `from_fetch` provenance flag, and the 429 `retry-after` hint that
 /// [`apply_outcome`] turns into a deferred next-fetch slot.
+///
+/// One exception to "rotate only on 401": the ACTIVE, Keychain-installed
+/// profile rotates ahead of expiry (see [`ACTIVE_ROTATE_LEAD_MS`]) so the
```
uwuclxdy: `[ACTIVE_ROTATE_LEAD_MS]` no longer exists; the v2 rename split it into `active_rotate_lead_ms` (the function) and `ACTIVE_ROTATE_LEAD_FLOOR_MS` (the constant). rustdoc treats this as an intra-doc link and cannot resolve it. CI stays green only because `rustdoc::broken_intra_doc_links` is not denied and doctests are self-gated. Please point it at `active_rotate_lead_ms`.

src/oauth.rs
```diff
@@ -642,12 +647,204 @@ pub(crate) fn apply_rotated_tokens_locked(
...
+        // the single-use refresh token the macOS Keychain copy carries — the
+        // running `claude` (which re-reads the Keychain per request) would
+        // sign out at that stale token's expiry while every clauth copy stays
+        // green (observed on-device 2026-07-07). Mirror the fresh pair into
+        // the Keychain inside the same locked section as the persist, so a
+        // concurrent switch cannot re-point the item in between. A mirror
+        // failure is loud but non-fatal: the rotation itself is durable, and
+        // the next rotation or switch retries the write.
+        #[cfg(target_os = "macos")]
+        if crate::keychain::enabled() && cfg.is_active(name) {
```
uwuclxdy, line 683: This runs `keychain_write` inside the `with_state_lock` closure, so the `/usr/bin/security` shell-out (up to its 20s kill deadline) holds the cross-process state flock the whole time. A hung `security` child would block every other clauth instance's state writes for those 20 seconds; before this PR the locked section contained only fast disk writes. The switch-repoint race being guarded only needs the fresh `creds` captured under the lock, not the subprocess call itself. Please capture `creds` under the flock and write the Keychain after releasing it.

src/oauth.rs
```diff
@@ -642,12 +647,204 @@ pub(crate) fn apply_rotated_tokens_locked(
...
+/// On success the mirror uuid is cached (`ACCOUNT_ID_CACHE_FILE`), so later
+/// adopts can verify identity even when the stored token is already dead.
+/// The Keychain is NOT written here — in this state CC minted the pair, so
+/// the Keychain and mirror are already the fresh truth; only our store lags.
+///
+/// Returns the adopted `(access, refresh)` pair so the caller can sync its
+/// in-memory `TokenList` exactly like every other rotation site — without it,
+/// the next poll would run on the superseded entry, spend the revoked refresh
+/// token, and fail on the very account the adopt just saved.
+pub(crate) fn try_adopt_live_rotation(
```
uwuclxdy, line 753: This mutates the same stored credential fields that `rotate_one_inner` only touches under the per-profile `RotationGuard`; here it is the state flock alone. Two writers using different locks on one resource is fragile against future refactors. Please take the guard here as well, or add a comment explaining why the flock alone is sufficient.

tests/inline/oauth.rs
```diff
@@ -309,3 +309,255 @@ fn rotation_guard_is_independent_across_profiles() {
...
+}
+
+// ── try_adopt_live_rotation (rotation coherence, the adopt-don't-race half) ──
+//
+// The running claude and clauth hold ONE single-use refresh family; when CC
+// rotates first, its file mirror (~/.claude/.credentials.json) carries the
+// fresher pair. Adopting it — identity-guarded — replaces racing for the
+// chain. All offline: identity is injected, the "mirror" is a sandboxed file.
+#[cfg(target_os = "macos")]
+mod adopt_live_rotation {
```
uwuclxdy, line 394: `mod adopt_live_rotation` is `#[cfg(target_os = "macos")]`, but `try_adopt_live_rotation` is platform-independent (only the `keychain_live()` call site is macOS-gated, not the function). As a result the identity gate and the expiry-monotonicity re-check, the correctness core of the adopt path, never run on the Linux leg or a Linux dev box. Please un-gate this module, keeping `keychain_mirror_gate` macOS-only since it references the `cfg(macos)` `live_login_is_foreign`.

uwuclxdy (changes requested):

xingfanxia:
All six addressed in `95d6928` (+ `296216a`, a test-flake hardening):

- **`preemptive_rotation` toggle** — new `AppState` field, `#[serde(default, skip_serializing_if = "is_false")]` matching the existing toggles, off by default (absent from old state files → `false`; off is omitted from the rendered toml — serde contract pinned by `preemptive_rotation_defaults_false_and_round_trips`). Surfaced as a `rotation: lazy / preemptive` Config-tab row with a hint line, same persistence shape as burn-aware. `proactive_rotation_due` takes the flag as its first gate, so stock clauth is strictly lazy end to end (`preemptive_rotation_is_opt_in_and_off_by_default`).
- **README:73** — reframed around adoption (concede-and-adopt, identity-guarded), mirror-on-rotate as the write half, and the early lead described as the opt-in toggle, explicitly "an optimization that makes adopt events rarer, not a correctness mechanism". The ~30-min/race framing is gone.
- **Flock never spans the `security` shell-out** — the mirror decision + creds snapshot stay under the locked section (so the written pair is exactly the persisted one), `keychain_write` runs after release. One nuance worth stating: the in-lock write never actually serialized against a *cross-process* switch either — the `is_active` check reads the in-memory config, which doesn't see another process's switch — so capture-under-lock preserves the exact prior guarantee (in-process switches stay excluded by the config mutex held across the function) while dropping the 20 s stall. Comment in the function says as much.
- **Adopt under the rotation lock** — `try_adopt_live_rotation` now takes `&RotationGuard` as proof-of-lock (the flock isn't reentrant, and the refresh-failure retry site already holds it — a parameter can't deadlock where an internal acquire would). The scheduler acquires the guard at rotation-leg entry, ahead of the adopt, matching `rotate_one_inner`'s guard-outermost ordering. Side effect worth noting: if the guard is busy, the adopt is now skipped for that tick (previously it ran unguarded) — the next tick retries, and the lock holder is precisely someone rotating/starting this profile.
- **`mod adopt_live_rotation` un-gated** — the identity gate + expiry-monotonicity re-check now run on every OS; each test acquires the guard it must now present. `keychain_mirror_gate` stays macOS-only.
- **`scheduler.rs:321`** — points at `active_rotate_lead_ms`.

611 tests (×many — also hardened `has_live_session_true_when_any_session_alive` with the same settled-poll as `live_session_count`, it reproduced ~1-in-10 during verification), clippy + rustfmt clean.

uwuclxdy:
Verified all six against 296216a. Everything checks out:

- `preemptive_rotation`: serde contract, default-off gate, Config-tab row, round-trip tests all confirmed end to end.
- The README bullet reads correctly now; the adopt framing matches the code.
- Mirror write: decision plus snapshot under the locked section, `security` shell-out after release. Your note about the cross-process `is_active` read matches what I see. Keeping the config mutex held for the whole function to exclude in-process switches is the right call.
- Adopt under the guard: acquiring at rotation-leg entry keeps the happy-path fetch flock-free; both adopt sites present the guard and the lock ordering vs config/state holds. The skip-when-busy tick is fine; the holder is rotating this profile anyway.
- Un-gated adopt tests and the doc link: confirmed.

Nothing further needed on this branch. #25 lands first (both PRs rewrite the same rotation leg), so expect a rebase over it; I'll do the final pass on the rebased branch and merge then.

xingfanxia:
Merged `mommy` (v0.9.0) into this branch to clear the conflicts from the release — resolutions: README auto-switch bullet takes the post-#26 weekly-cap wording, `usage/mod.rs` re-export list is the union. No behavior change; 647 tests green.

uwuclxdy:
Merged via a local merge commit (b0a1f33), with the rebase reconcile done on my side: `fetch_with_rotation` is now the union of this PR's guard-at-entry + adopt + preemptive leg with #25's 429-unmask + carry machinery. Two seam items landed as a follow-up (1794812). The adopt path lifts a stale `auth_broken`, the line your v2 comment predicted for whichever PR landed second; that pick-up fell to me. The proactive check reads the stored expiry from config again rather than the entry snapshot, which goes stale after a same-tick kick rotation. Thanks for both PRs; the rotation story is in much better shape.

PR #25, opened 2026-07-08
feat(auth): unmask dead logins behind 429s, quarantine auth_broken accounts

xingfanxia:
This is offer 1 from #1, cut as the design-in-code you can review directly — the shape is exactly what I described there, and I'm happy to rework any part of it.

## Why

A dead account (refresh token revoked / login silently dropped) often comes back **429, not 401** — and a poller that treats every 429 as "endpoint busy, serve cache" hides a genuinely dead login forever. On my fork this is the failure mode that let dead accounts sit in the fallback chain looking healthy; cross-platform value, since the 429 costume hides dead logins on Linux too.

## What

**429 unmasking.** A 429 on a token that is *also* past its clock expiry falls through to the rotation leg instead of backing off (`token_clock_expired`) — it would 401 the moment the limit cleared anyway, so the 429 is masking a token that must be refreshed. A 429 on a still-valid token bails to cache exactly as today: never spend the single-use refresh on a real endpoint rate limit. Unknown expiry never rotates (conservative).

**Terminal-vs-transient refresh.** New `refresh_result` preserves the distinction (`RefreshError::Invalid` for 400/401/403 from the token endpoint = revoked; `Transient` for transport/429/5xx). The existing `refresh()` keeps its `anyhow` signature — current callers are untouched.

**Quarantine.** A terminally-failed refresh flags the profile in a persisted `AppState::auth_broken` list (logged once, on the transition). Quarantined members are excluded from every chain walk — `next_target` and `next_auto_switch_target` (via `ChainSnapshot`) — and refused as a CLI switch target through a pre-install gate (`ensure_installable`), because installing a dead token into the macOS Keychain signs out every running `claude`. The gate also refreshes a merely *expiring* target before install, so a fresh switch never lands on an immediate 401. Transient failures refuse the one attempt and never quarantine a healthy account.

**Recovery.** A successful refresh or `clauth login <name>` (fresh capture or your #22 in-place reauth) clears the flag. Delete/rename hygiene keeps the list consistent.

## Coverage notes (honest edges)

- The pre-install refresh gate covers the CLI switch (`switch_profile_cli`), which owns a `ConfigHandle` and can drop the lock across the refresh HTTP. The TUI's in-app switch calls `switch_profile(&mut AppConfig)` under the app lock, so it can't host the gate as-is — it's protected by the walk exclusion (a dead account gets quarantined by the poller and is never picked again) plus the normal 401→rotate path. If you want the gate on the TUI path too I'm happy to follow up, but it needs a call-site restructure I didn't want to smuggle into this PR.
- While touching the token-endpoint parse errors: a 2xx body that fails to deserialize still holds live tokens, so parse failures now report only category/position/status/length (`token_parse_error`) instead of echoing the body — applied to the new `refresh_result` and the pre-existing `exchange_code` site.

## Tests

606 total (13 new): the `ensure_installable` gate ×6 (fixture refresher, offline — third-party bypass, valid-token ready, expiring-refreshes, invalid-quarantines, transient-doesn't, no-refresh-token-broken), walk exclusion, the two 429 classifiers, reauth-clears-flag, and the parse-redaction lock. Clippy + rustfmt clean. No real Keychain or credentials are touched anywhere in the suite.

Heads-up: overlaps with #24 in `fetch_with_rotation` — happy to rebase whichever lands second.


uwuclxdy:
thanks!

could you just turn the verbose comments / comments that assert behavior into tests instead? you can add short explanatory comments there if needed.

xingfanxia:
Done in 3e23619 — the behavior-asserting prose now lives in tests, comments keep only the WHY (with a pointer to the pinning test where that helps a reader):

New tests (6, suite 606 → 612):
- `set_auth_broken_reports_transitions_and_is_idempotent` — the transition-only contract that gates the one-line log (no per-tick spam).
- `auth_broken_survives_a_state_save_load_round_trip` and `state_file_without_auth_broken_field_loads_clean` — persistence + old-state-file tolerance (`serde(default)`).
- `remove_profile_drops_its_auth_broken_entry` — delete hygiene.
- `walk_excludes_broken_members_from_the_last_resort_pass_too` — the exclusion covers BOTH walk passes, not just headroom.
- `cli_switch_refusal_names_the_login_recovery` — the Broken-gate error actually tells the user the fix (`clauth login <name>`), driven through a fixture refresher.

Shrunk: the `ensure_installable` bullet-list doc, the two 429 classifier docs, `mark_auth_broken`'s transition rationale, `AppState::auth_broken`'s field doc, and the parse-redaction doc — each now 3-6 lines of rationale instead of full behavioral prose. Clippy + rustfmt clean.

xingfanxia:
Two more commits from a live incident yesterday (2026-07-09) that this PR's quarantine machinery exposed:

**`7ecdda8` — AUTH-4: a dead login is itself a switch trigger.** Once an active profile is flagged `auth_broken`, it can never produce a `Fresh` usage read again, and its frozen store entry is typically a lapsed 5h window that reads as idle headroom. Both auto-switch gates (`scan_auto_switch`'s freshness gate and `next_auto_switch_target`'s exhaustion gate) therefore stall forever, and the daemon sits on the dead account while a viable sibling idles — we watched it do exactly that for several hours. Fix: a broken active bypasses both gates and drives the walk; the wrap-off (halt-all) leg deliberately still keys on **real exhaustion only**, so the flag alone never logs out a live session over quota it hasn't spent. The TUI-side `auto_switch_if_needed` gets the same parity guard. Tests are mutation-verified — reverting either gate fails them.

**`82a6354`** — same test hardening as on #24: `live_session_count_counts_only_alive` polls until settled instead of asserting one reading (the fail-alive probe flakes ~1-in-3 under parallel load; it reproduced here during verification).

618 tests ×4, clippy + fmt clean. Note both this branch and #24 now carry the identical runtime-test hunk — whichever lands second will merge it trivially.

src/usage/scheduler.rs
```diff
@@ -306,11 +332,21 @@ fn fetch_with_rotation(
...
-    let tok = match refresh_result {
+    let tok = match rotation {
         Ok(t) => t,
-        Err(_) => return bail_to_cache(None),
+        Err(e) => {
+            // A terminal failure (dead refresh token) quarantines the account on
+            // this tick; a transient one leaves the flag and retries. See
+            // `refresh_failure_is_terminal`.
+            if refresh_failure_is_terminal(&e) {
+                crate::oauth::mark_auth_broken(config, name, true);
```
uwuclxdy, line 402: A single 400 here quarantines terminally, but "refresh token not found or invalid" is also the exact response after a benign single-use double-spend: CC refreshing the active profile's symlinked credentials mid-poll, or a second `clauth` instance racing this one. `has_live_session` only guards `clauth start` sessions, so a healthy active profile can end up quarantined, after which AUTH-4 walks the daemon off it. On a 400, please re-read the on-disk credentials under the guard first: changed credentials mean another refresher already rotated, so retry with the fresh pair. Only an unchanged-credentials 400 should set `auth_broken`.

src/oauth.rs
```diff
@@ -69,28 +92,66 @@ static AGENT: LazyLock<ureq::Agent> = LazyLock::new(|| {
...
         .send(&body)
-        .map_err(anyhow::Error::from)?;
+        .map_err(|e| RefreshError::Transient(anyhow::Error::from(e)))?;
     let status = response.status().as_u16();
     let text = response
         .body_mut()
         .read_to_string()
-        .map_err(anyhow::Error::from)?;
+        .map_err(|e| RefreshError::Transient(anyhow::Error::from(e)))?;
+    if matches!(status, 400 | 401 | 403) {
```
uwuclxdy: Routing 403 into permanent `Invalid` means one transient WAF/geo/challenge block quarantines a healthy account until its next successful refresh. A revoked refresh token comes back as a 400 `invalid_grant` (or a 401) per OAuth2. Please keep 400/401 terminal; route 403 to `Transient` unless the response body actually confirms `invalid_grant`.

src/actions.rs
```diff
@@ -130,6 +130,22 @@ pub(crate) fn switch_profile_cli(config: AppConfig, canonical: &str) -> Result<(
 
     let config = Arc::new(RankedMutex::new(config));
 
+    // AUTH-1 (Incident C): gate the target before its credentials land in the
+    // Keychain (which re-authenticates every running `claude` on this machine).
+    // Refusal + `clauth login` hint pinned by
+    // `switch_cli_refuses_dead_target_with_login_hint`.
+    match oauth::ensure_installable(&config, canonical, oauth::refresh_result) {
```
uwuclxdy, line 137: `ensure_installable` gates only this CLI path. The MCP `switch` and the TUI switch go through the noninteractive path with no gate, so selecting a quarantined profile there still installs the dead token into the Keychain, the Incident C scenario the README bullet says is prevented. Please wire the gate in there as well so "refused as a switch target" holds for every entry point.

uwuclxdy (changes requested):
The token-safety core is correct: the clock-expired-429 rotation reuses `RotationGuard` with the in-guard `has_live_session` skip. The clock-expired state self-clears after one successful refresh, so there is no per-tick token burn. Extending the fetch path past the previous "the poll never rotates on a 429" rule is fine; I will reconcile my own notes at merge. Three requests:

- The terminal 400 needs a double-spend guard. "refresh token not found or invalid" is also what a benign single-use double-spend returns: CC refreshing the active profile's symlinked credentials mid-poll does exactly that (the `rotate_one_inner` comment documents it). `has_live_session` only covers `clauth start` sessions. On a 400, please re-read the on-disk credentials under the guard: if they changed, another refresher already rotated, so retry with the fresh pair instead of quarantining. Only an unchanged-credentials 400 is a real revocation.
- 403 should not be terminal on its own. A WAF/geo/challenge 403 would quarantine a healthy account until its next successful refresh. Please route 403 to `Transient` unless the body confirms `invalid_grant`; 400/401 stay terminal.
- Please gate every switch entry point. `ensure_installable` has a single call site (`switch_profile_cli`), so the MCP `switch` and the TUI switch can still install a quarantined token into the Keychain, the exact Incident C scenario that `README.md:73` says is prevented. Wiring the gate into the noninteractive switch path makes the README claim hold as written.

Happy to discuss the re-read shape; the other two should be mechanical. Thanks!

xingfanxia:
All three addressed in `ef75932` (+ `43e0887` flake hardening, `5f6edd6` a self-review catch on the new path):

- **Terminal-400 double-spend guard** — on a terminal failure the poll re-reads the profile's on-disk credentials while still holding the rotation guard (`fresher_disk_pair`): a stored pair that moved past the token we just spent means another refresher already rotated — CC through the symlinked live file, or one that completed before this tick's guard was acquired (the in-memory `TokenEntry` snapshot predates the guard, which is the race the guard alone can't close). The fresh pair is carried into the `TokenList` and refetched next tick; only an unchanged-credentials 400 quarantines. Truth table for the helper pinned offline. Follow-up caught in pre-push self-review (`5f6edd6`): the carry must also **lift** a stale quarantine — the moved pair proves the chain is alive, and without the clear a recovered account stayed flagged forever, since the carry preempts the refresh that would normally clear it (a wrong lift self-corrects next tick: the carried pair's own 400 with an unchanged store re-quarantines).
- **403 → Transient unless confirmed** — `refresh_rejection_is_terminal` (extracted pure, truth table pinned): 400/401 terminal regardless of body; 403 terminal only with `invalid_grant` in the body; WAF/geo/challenge 403s route to Transient and never quarantine.
- **The gate covers every entry point** — `switch_profile_noninteractive` now takes the shared `ConfigHandle` + an injected refresher and runs `ensure_installable` before any relink, so the MCP `switch` (and any future headless caller) shares the CLI's exact refusal, `clauth login <name>` hint included; pinned by `noninteractive_switch_refuses_a_dead_target_with_login_hint` (+ a Transient-arm test proving no quarantine on a blip). The TUI switch gets a synchronous flag-only refusal (`finalize_switch` runs on the UI thread — no HTTP there; the flag is kept current by the poller and every refresh site), pinned by `tui_switch_refuses_a_quarantined_target_with_login_hint`. Your README line now holds as written.

One place deliberately NOT given the disk re-read: `ensure_installable` itself. Its refresh runs under the rotation guard against a profile that is normally not the ACTIVE one (a switch *target*; switching to the already-active profile no-ops earlier), so the CC-mid-poll racer doesn't exist there, and clauth-side racers are excluded by the guard. Happy to add it if you'd rather have symmetry.

625 tests (×8 full-suite runs on the pre-5f6edd6 tree, once after) clean, clippy + rustfmt clean.

uwuclxdy:
Confirmed all three on 5f6edd6:

- The double-spend guard reads the on-disk pair back under the same held rotation guard. The refresh-token comparison is the right discriminator for a single-use chain. The 5f6edd6 lift is correctly bounded: a dead chain never advances the disk pair, so a wrong lift re-quarantines on the next tick.
- `refresh_rejection_is_terminal` matches the ask exactly, truth table included.
- The MCP path now runs the gate before any relink. The TUI flag-only refusal is a reasonable UI-thread tradeoff, documented as such.

Verifying the gate surfaced two things, both small:

1. `switch_profile_cli` and `switch_profile_noninteractive` run `ensure_installable` before anything checks whether the target is already active (the `is_active` no-op lives inside `switch_profile`, after the gate). A switch to the currently active profile therefore refreshes the active chain whenever it sits inside the grace window. That is the one profile a plain `claude` may be refreshing concurrently through the symlink; if CC wins that race, the gate's refresh 400s and `mark_auth_broken` false-quarantines a healthy active. The scheduler's carry lifts the flag a tick later, but the switch was already refused with a wrong "login expired" hint. The flag also excludes the profile from the fallback walk until the lift. Your note says the already-active case no-ops earlier; the call ordering says otherwise. Please return early on target == active before the gate in both entry points; there is nothing to install in that case.
2. `ensure_installable` never consults `auth_broken`; it gates on clock expiry alone. A quarantined profile whose access token is revoked server-side but still clock-valid passes the CLI and MCP gates as `Ready` and installs the dead token. The new TUI check refuses that same target, so the entry points currently disagree. Please treat a flagged profile as expiring so it always routes through the refresher: a recovered chain comes back `Refreshed` and lifts the flag, a dead one confirms `Broken` with the login hint.

Tiny: the doc comment on `no_network` in mcp_switch.rs says it is never invoked, but the Transient-arm test drives the gate through it via an expired target.

Both fixes should be a few lines each; the rest is mergeable as is.

xingfanxia:
Merged `mommy` (v0.9.0) into this branch to clear the release conflicts. The one substantive touch: `refresh_result` now routes both its error arms through the new `http_error()` body-capping helper, so the quarantine reason strings get the same first-line/200-char cap as everything else. README unions the 429-unmask doc with the post-#26 weekly-cap wording. 663 tests green.

uwuclxdy:
Merged. I took the two follow-ups myself in 99a9949: the switch paths now exempt the already-active profile from the gate, and `ensure_installable` routes a flagged profile through the refresher regardless of the stored clock (a recovered chain lifts the flag on the way through). The `no_network` comment is fixed up as well. Thanks for the quick turnarounds on this one.

PR #26, opened 2026-07-08
fix(fallback): weekly (7d) cap hard-blocks a chain member — both directions

xingfanxia:
Found on-device 2026-07-08, and upstream has the same hole (`fallback.rs` never reads `seven_day`): **auto-switch rotated INTO an account whose weekly window was spent to 100%.**

## Why the walk loves dead accounts

A weekly-capped account is the pathological case for 5h-only exhaustion logic: the API refuses its requests, so its 5h window drains and then **lapses entirely** (`resets_at: null`) — which every predicate reads as *fresh headroom*. On my machine two chain members sat at exactly `7d = 100.0` with no live 5h window; the walk targeted one, and `find_recovered_member` re-linked weekly-dead members on every 5h rollover ("5h lapsed = recovered").

## Fix

A LIVE overall `seven_day` window at ≥ 100% (`WEEKLY_BLOCK_PCT`) counts the account exhausted in every predicate — profile-side and store-side, the target walk and the ACTIVE side (where the hard block precedes the burn-aware projection: there is nothing left to project), and recovery never picks a weekly-dead member. `soonest_resume` reports the WEEKLY reset for such members instead of bailing on their absent 5h window, so the all-exhausted caption stays truthful.

Deliberately a **hard-block check, not a threshold**: 99.9% still serves requests, and a 7d=100 whose reset has passed is a renewed quota. Only the overall `seven_day` window gates — a per-model `weekly_scoped` limit at 100 (e.g. a model-scoped weekly spent while 7d has room) blocks just that model, and the walk can't know which model the next session will drive.

## Tests

595 total (6 new): never-a-target, active-side exhaustion despite an idle 5h window, below-cap/lapsed-reset headroom, the store-side walk twin, never-recovers (+ recovers once the weekly reset passes), and the soonest-resume caption. Clippy + rustfmt clean.

Heads-up: overlaps #24/#25 in `fallback.rs`/`scheduler.rs` context — happy to rebase in whatever order you land them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Most account tools do one half. clauth pairs instant **switching between multipl
 ### Automate & stay safe
 
 - **Automatic token refresh**: OAuth refresh tokens are single-use, so rotation stays lazy. A stale access token rotates the moment a usage query 401s, never ahead of time. <kbd>t</kbd> force-rotates every account.
-- **Auto-switch on exhaustion**: opt accounts into an ordered fallback chain. When the active one crosses its 5h threshold (95% default), clauth hops to the next member with headroom. An opt-in burn-aware mode (Config tab) switches on projected usage instead: heavy burn hops early, light burn rides closer to 100% before moving. Needs clauth open.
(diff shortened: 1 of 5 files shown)


uwuclxdy (approved):
The weekly hard-block reaches every fallback predicate: profile and store variants, target and active walks, recovery, `soonest_resume`. The burn-off paths still match `is_exhausted`, with the store lock dropped before the burn read. The wall-clock rule holds as well: `seven_day_live` gates on a parseable future reset, so a fully-used weekly window whose reset already passed reads as renewed rather than blocked. The tests assert exact values in both directions and fail if the gate is removed. Two doc lines went slightly stale (`soonest_resume`'s header still describes the 5h-only bail; the README "never a switch target" line is unqualified for the `last_resort` sink). I will fix both at merge, thanks!

PR #32, opened 2026-07-12
fix(tui): fixed_split silently drops one char at the truncation boundary

xingfanxia:
## The bug

`fixed_split`'s `for ch in iter.by_ref()` checks `count >= width` at the top of the body — **after** the iterator has already yielded the char. So for a value exactly one char over the window, the overflow char is consumed by the loop head, the follow-up `iter.next()` sees an exhausted iterator, and no ellipsis is added: the cell renders `width` chars with one char silently cropped.

Concrete: the overview's `type` column at its 6-wide band renders `"Max 20x"` as **`Max 20`** — a plausible-but-wrong tier (reads as Max 20, is Max 20x) — instead of `Max 2…`. Same for any name/label/email exactly one char over its column: you get a valid-looking wrong string with no truncation cue, which is worse than an honest ellipsis.

## The fix

Peek instead of consume: a `while count < width` loop pulls only chars inside the window, and the ellipsis check uses `iter.peek()`. Any dropped character now always surfaces as `…`.

## Tests

New `tests/inline/tui_render_format.rs`:
- the boundary case (`"Max 20x"` @ 6 → `"Max 2…"`, and a 16-char email @ 15),
- exact-fit / shorter / far-over / width-0 behavior unchanged,
- a small property sweep pinning `content + pad` at exactly `width` chars for value lengths 0–12 × widths 1–9.

709 tests green on `mommy`, clippy/fmt clean.

### DIFF
--- a/src/tui/render/format.rs
+++ b/src/tui/render/format.rs
@@ -20,15 +20,17 @@ pub(super) fn fixed(value: &str, width: usize) -> String {
 pub(super) fn fixed_split(value: &str, width: usize) -> (String, String) {
     let mut content = String::with_capacity(width);
     let mut count = 0;
-    let mut iter = value.chars();
-    for ch in iter.by_ref() {
-        if count >= width {
-            break;
-        }
+    // Peek instead of a plain `for`: the loop head must not consume the char
+    // AFTER the window, or a value exactly one char over `width` reads as
+    // exhausted and gets silently cropped with no ellipsis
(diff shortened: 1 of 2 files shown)


uwuclxdy:
Merged, thanks! the `Max 20x` → `Max 20` repro plus a test that goes red on `mommy` first is exactly the shape i want. landed as-is via rebase.

Issue #27, opened 2026-07-08
discussion: headless daemon + status feed — should any of it live upstream?


xingfanxia:
Splitting offer 2 from #1 into its own thread so it doesn't tangle with the auth work. This is a **should-we / how-much** discussion, not a PR — the surface is big and it's your call how much of it belongs in clauth.

## What exists (running on my fork for ~2 weeks)

**`clauth daemon`** — the usage-refresh + auto-switch loop with no TUI:
- Same scheduler/walk code the TUI runs (shared modules, not a fork of the logic); auto-switch, burn-aware projection, and the fallback chain all work unattended. Your README's "Needs clauth open" caveat on auto-switch disappears.
- Writes `~/.clauth/status.json` every tick — a stable JSON snapshot (schema-versioned) of profiles, 5h/7d windows, chain order + thresholds, active/pending switch, last error, and the daemon's own next-move **forecast** computed by the same `next_target` walk the switch decision runs (published precisely so no client ever re-implements the walk and drifts).
- Serves `~/.clauth/clauthd.sock` — line-JSON commands: snapshot / switch / refresh / chain config (add/remove/move/threshold/last-resort/wrap-off) / rename.
- Single-instance lock, config hot-reload on external changes, log rotation, launchd packaging for macOS login start.

**[ccsbar](https://github.com/xingfanxia/ccsbar)** — a native macOS menu-bar app consuming that feed (usage bars, forecast sentence, one-click switch, login-health badges with browser reauth, in-app add-account via `clauth login --new`). Stays a separate repo regardless — it's macOS-only Swift and clearly not clauth's problem; it matters here only as proof the feed contract works and as its consumer.

## Questions for you

1. **Appetite**: do you want a headless mode in clauth at all, or would you rather it stay fork-side? Zero pressure — the fork is happy to carry it, but issue #1's discussion suggested the burn-aware + unattended combination has cross-platform value (a Linux user can run it under systemd just as well; nothing in the daemon core is macOS-specific).
2. **If yes, how much?** Options by increasing surface:
   a. Just `status.json` + `clauth status --json` (the read-side contract; any tray app on any OS can build on it);
   b. (a) + the daemon loop (unattended auto-switch, no socket);
   c. (b) + the control socket (full remote-control surface).
3. **Contract review**: if any of it comes over, the JSON shapes and socket protocol deserve your review *before* a PR — happy to write them up as a spec first so the review is about the contract, not 2k lines of Rust.

[...]

PR #34, opened 2026-07-12
fix(auth): ensure_installable re-reads token state under the RotationGuard

xingfanxia:
Follow-up to #25 — this is the "pre-`RotationGuard` token-snapshot race in the `ensure_installable` gate" you flagged in the #27 review.

## The race

`ensure_installable` snapshots `(expires_at, refresh_token, flagged)` under the config lock, releases it, and only **then** acquires the `RotationGuard`. In that window a sibling rotation (a live session's chain advance, the scheduler's `rotate_one`, another switch path) can spend the single-use refresh token and persist a new pair. The gate then feeds the **stale snapshot** to the refresher:

- the spend 400s as `Invalid` → `mark_auth_broken(name, true)` — a healthy login with a perfectly fresh pair on disk gets **wrongly quarantined**, and
- depending on server-side rotation policy, spending an already-consumed token can burn the chain that just got minted.

## The fix

The refresh leg is split into `gate_under_guard`, entered only with the guard held, and it **takes no token arguments** — post-guard decisions structurally cannot reuse pre-guard data:

- it re-reads `(expires_at, refresh_token, flagged)` under the guard (authoritative: the guard serializes rotations, so nothing can invalidate this read),
- a sibling-refreshed pair (no longer expiring, flag lifted) returns `Ready` and installs as-is instead of double-spending the old chain,
- still-expiring state refreshes from the **currently stored** token.

The guard-less pre-check stays as the cheap early-exit for non-OAuth / comfortably-live tokens; its token data is discarded. The `auth_broken`-overrides-the-clock semantics from the quarantine work are preserved verbatim in both legs (shared `expiring(expires_at, flagged)` helper).

## Tests

Two new tests pin the boundary directly on `gate_under_guard` (the function boundary is what makes the race unrepresentable, so the tests target it):
- `gate_under_guard_installs_a_sibling_refreshed_pair_as_is` — fresh stored pair → `Ready`, refresher (a panic fixture) never runs;
- `gate_under_guard_spends_the_currently_stored_refresh_token` — still expiring → the refresher receives the token read under the guard.

All 7 existing `gate_*` tests unchanged and green (including `gate_flagged_profile_refreshes_despite_a_valid_clock`). 705 tests, clippy/fmt clean.

### DIFF
--- a/src/oauth.rs
+++ b/src/oauth.rs
@@ -1017,40 +1017,16 @@ pub(crate) fn ensure_installable(
     name: &str,
     refresher: impl Fn(&str) -> std::result::Result<TokenResponse, RefreshError>,
 ) -> AuthGate {
-    // Read the target's auth shape under the config lock, then release it — the
-    // HTTP refresh below must never hold the config mutex (a slow refresh would
-    // otherwise wedge the daemon's single-threaded run loop).
-    let (expires_at, refresh_token, flagged) = {
-        let Ok(cfg) = config.lock() else {
-            return AuthGate::Transient(anyhow::anyhow!("config mutex poisoned"));
-        };
-        let Some(profile) = cfg.find(name) else {
-            // Unknown profile: nothing to gate — the switch itself surfaces
-            // "Profile not found".
-            return AuthGate::Ready;
-        };
-        if !profile.is_oauth() {
-            // Third-party (api-key) profiles carry no OAuth token to expire.
+    // Cheap pre-check WITHOUT the rotation guard: non-OAuth and
+    // comfortably-live tokens install as-is. Token data read here is
+    // discarded — only the post-guard re-read may feed the refresher (a
+    // pre-guard snapshot can go stale the moment a sibling rotation runs).
+    match oauth_shape(config, name) {
+        Err(gate) => return gate,
+        Ok((expires_at, _, flagged)) if !expiring(expires_at, flagged) => {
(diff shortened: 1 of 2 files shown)


uwuclxdy:
Thanks for the PR! The refactor itself is sound: lock order `Rotation` then `config` is correct, config is dropped before the HTTP refresh, `expiring()` matches the old inline check byte for byte.

One gap tho. The re-read under the guard hits the in-memory `ConfigHandle` (`cfg.find(name)`), so it only observes rotations from this process (the TUI + its in-process scheduler share the `Arc`). `switch_profile_cli` and the MCP switch load config from disk once and never reload. So the cross-process case (a daemon rotates + persists between their disk load and the guard acquire) still refreshes the already-spent token, gets the terminal 400, `mark_auth_broken` quarantines a healthy login. That's the failure mode the PR body describes, still open for the primary callers. The guard serializes rotations; it doesn't propagate a peer's disk write into this process's memory.

Two smaller things:

- the "entered only with the `RotationGuard` held" doc comment is an unenforced contract; taking `_guard: &RotationGuard` as a witness param makes the compiler prove it.
- the two new tests duplicate existing `gate_*` coverage; neither exercises the interleave they're named for.

I'll push a fix to this branch (disk reload under the guard + the witness param), done with claude. lmk if the in-memory read was intentional for some reason i'm missing.

PR #35, opened 2026-07-12
feat(daemon): headless scheduler + status.json feed, TUI stand-down (#27 a+b)

xingfanxia:
The (a)+(b) cut you accepted on [#27](https://github.com/uwuclxdy/clauth/issues/27), built to the contract [posted there](https://github.com/uwuclxdy/clauth/issues/27#issuecomment-4951220501) — that contract now lives in the diff as `docs/daemon.md`. **Opening as a draft**: you asked for the spec review first, and the spec is still open on #27 — this gives you the implementation to read beside it, and it flips to ready (with any field renames/cuts applied) once the contract is agreed.

## What ships

- **`clauth daemon`** — runs the exact same `spawn_refresher` the TUI runs (no `if daemon` branches in `scheduler.rs`). Each 1 s tick: reload external config changes, execute queued auto-switches / wrap-offs, atomically rewrite `~/.clauth/status.json` (`0600`, tmp+rename, token-free). Singleton advisory lock with standby takeover; anti-wedge watchdog (a stuck tick aborts for a clean supervisor restart); size-capped log; `~/.clauth` tightened to 0700 on boot.
- **`clauth status --json`** — the same serializer single-shot from the on-disk caches, no daemon, no network. One code path builds both, so the feed and the CLI snapshot cannot drift.
- **Dual-scheduler dedup** — your condition for this landing. The TUI probes the singleton flock + feed freshness every tick and stands its own refresher down while a daemon is live: it hydrates its stores from the disk caches the daemon keeps fresh instead of double-polling the usage API, double-rotating single-use chains, or re-deciding switches — and re-arms within one tick of the daemon dying (flock released) or wedging (stale `generated_at`). The 30 s staleness window deliberately undercuts the 60 s watchdog: a wedged daemon gives a bounded ~30 s dual-fetcher overlap, which is merely duplicative and beats the refresh coverage gap of waiting the watchdog out. The daemon never probes (its own held flock would read as "another daemon"), and the transition logs stay off a live terminal — the TUI's stderr is the ratatui screen.
- **Timestamped daemon logging** — every daemon-visible stderr line gets an ISO-8601 UTC prefix, enabled only in daemon mode; interactive stderr stays bare.

## Scope held per your review

- **No socket, no external mutation surface** — (c) stays fork-side. Anything needing eyes (diverged live login, manual `/login`, unprovable identity) is refused and logged; the TUI remains the only resolution surface.
- `status.json` emits **exactly the posted contract** — schema 1, additive evolution. The fork's extra fields (`clauth_version`, `last_switch`, `last_error`, `forecast`, `burn_aware`, `fallback_chain`, `account_email`, `fallback.last_resort`) are deliberately NOT in this cut; each is additive under the evolution rule, and I'm happy to bring any of them in individually if you want them.
- Your two queued nits are addressed in this cut: the watchdog spawn failure is loudly logged (was `.ok()`-swallowed), and the never-set `shutting_down` is documented as deliberate at its construction site (process exit IS the daemon's shutdown under a supervisor; the flag's writer is the TUI quit path).
- `.agent/`, `docs/ccsbar/`, `media/` are not in the diff; the contract rides as a lean `docs/daemon.md`.

## Adaptations made during the port (vs the fork)

- The fork's ordered pending-switch queue (origin precedence, per-entry TTL) collapsed onto your `HashSet<String>` `PendingSwitch`: without a socket there is only ever the scheduler's own target, so the drain takes the set atomically and the 120 s give-up window moved into the per-target backoff state. Failed switches retry on a deduped exponential backoff — one log line per distinct reason, never 1/tick.
- `collect_tokens` moved from `tui/app.rs` into the scheduler and takes `&AppConfig`, so the TUI and the daemon share one definition (it snapshots `auth_broken` for the poll partition either way).
- The mcp server's private provider/tier/windows JSON helpers consolidated into the new `profile_json` module the daemon serializer uses — the three JSON surfaces (mcp, `status.json`, `status --json`) genuinely share one shape now.
- The per-target give-up TTL (120 s) is enforced at the backoff gate too, so a capped backoff step near the window's edge can't outlive it.

## Verification

- 734 tests green, `cargo clippy --all-targets` clean, `cargo +nightly fmt` clean.
- New coverage: daemon drain characterization (idle no-op, executes-when-clean, divergence skip, busy-target requeue, backoff dedup over many ticks, TTL closes at the window edge, own-write mtime adoption + external reload), status.json shape with exact key-set pins + auth_status + pending_switch, probe truth table (flock held/unheld × feed fresh/stale/garbage), stand-down hydrate contract, log rotation, watchdog decision, 0700 migration.
- Running live on the fork's deployment (macOS launchd, `KeepAlive{SuccessfulExit=false}`) for several weeks in earlier form; this exact cut is what the fork now runs modulo the fork-side extensions.


xingfanxia:
Rebased onto current `mommy` (v0.9.1 + api-key login + profile delete) — clean replay, 757 tests green, clippy clean.

One new commit worth your eyes: **`clauth delete` × the daemon's pending-switch queue**. The queue holds a raw profile name in this process's memory, so an out-of-process delete can't purge it; pre-fix, draining that ghost ran `switch_profile`, whose existence check fired only in `finish_switch` — *after* `force_link` had torn down the live credentials link — and the retry then misread the missing live file as "logged out" and nulled the ACTIVE profile's stored credentials. Two guards now in the diff: `switch_profile` validates existence before any side effect, and the drain drops a vanished target with an accurate logline instead of retrying it. Characterization test included (`drain_pending_switch_drops_a_vanished_target`). This interaction only exists once the daemon lands beside your new `delete`, which is why it rides this PR.

Also widened the fail-alive liveness-probe poll windows in the runtime tests 2s→10s (one flaked under a parallel full-suite run; the probe reads transient try_lock I/O errors as "alive" by design, so the tests must out-wait fd-pressure transients).

Spec on #27 unchanged — still draft pending your review of the `status.json` contract.

xingfanxia:
Heads-up: the `switch_profile` existence-first guard that rides in this PR (`src/actions.rs` hunk of the vanished-target fix) is now also extracted as a standalone PR — #36 — since the hazard is reachable on `mommy` today (v0.9.1 `clauth delete` × the async switch gate) and should not wait on the #27 spec discussion. If #36 merges first, I will rebase this branch and the duplicate hunk drops out cleanly; the daemon-side half (`drain_pending_switch` dropping a vanished queued target) stays here. Also opened from the same batch: #37 (active-profile 429 ladder cap) and #38 (weekly soft line, configurable) — all three have been running in production on the fork.

uwuclxdy:
Verified the parts I care most about before anything else:

- no double rotation even in the stand-down overlap window: the cross-process `RotationGuard` flock + the `fresher_disk_pair` carry hold
- lock order is clean (`StateLock` enters the rank only at depth 0, so the daemon's nested `with_state_lock` is fine)
- crash durability checks out: watchdog abort, atomic `0600` writes, dual-signal stale-daemon detection
- scope held to #27 a+b

Findings to settle before external readers build against the feed:

- third-party (api-key) profiles always emit `fetch_status: null` + empty `windows`: `mtime_ms` reads the OAuth `USAGE_CACHE_FILE`, `LiveSignals.status` never contains an api-key name. Per the field contract null means "never fetched", so a healthy hourly-refreshed Z.ai/DeepSeek profile renders permanently as never-fetched. Freshness should come from `THIRD_PARTY_CACHE_FILE` mtime for those.
- `docs/daemon.md` examples use the `Z` suffix, `epoch_secs_to_iso` emits `+00:00`. A reader keying strictly on `Z` misparses the real feed.
- the log-rotate in-place trim is only sound for an `O_APPEND` fd. A systemd `StandardError=file:` redirect keeps the fd offset, so you get a sparse NUL hole + the size cap defeated. Worth a hard doc requirement since no unit file ships.
- `clauth status --json` can never report `Failed`/`RateLimited` (mtime-derived), the live feed can. The "cannot drift" claim in the doc holds for shape only, worth qualifying.
- profiles named exactly `daemon` or `status` get silently shadowed by the new subcommands.

I'll push fixes for the first two + the doc qualifications to this branch, done with claude. The reserved-name guard can be a follow-up. Planning to merge this one after the smaller PRs so you don't eat rebases on a 2.9k-line diff.

PR #36, opened 2026-07-12
fix(actions): a switch to a vanished profile bails before side effects

xingfanxia:
## The hazard

`clauth delete` (new in v0.9.1) can remove a profile that another switch surface still holds **by name**: a queued auto-switch target posted by the scheduler, the MCP switch tool, or a CLI switch racing the delete.

`switch_profile`'s existence check lived **last** in the sequence — inside `finish_switch` — so by the time a ghost target was noticed, `snapshot_active_credentials` + `force_link_profile_credentials` had already torn down the live `~/.claude/.credentials.json` link. The switch fails anyway, but only *after* destroying the live login: every running Claude Code gets logged out, and the half-mutated link state is left for the next divergence/classification pass to misread.

We hit the worst-case downstream of this on our fork (a retry loop misread the half-torn link as "logged out" and nulled the active profile's stored credentials); the unguarded window itself is in this repo's `switch_profile`.

## The fix

Validate existence **first**, before any side effect, for every caller:

```rust
if config.find(name).is_none() {
    bail!("profile '{name}' not found");
}
```

## Verification

- New test `switch_to_a_missing_profile_bails_before_touching_the_live_link` — RED against the old code (live link was destroyed), GREEN with the guard: the error surfaces and the live credentials file is byte-identical after the failed switch.
- Full suite: 727 passed, clippy clean.

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -56,6 +56,15 @@ pub(crate) fn validate_profile_name(
 
 pub(crate) fn switch_profile(config: &mut AppConfig, name: &str) -> Result<()> {
     with_state_lock(|| {
+        // Existence FIRST: everything below has side effects (the live
+        // credentials link is torn down before `finish_switch` would notice a
+        // ghost), and a caller holding a stale name — a queued auto-switch
+        // target, the MCP switch tool, or a CLI switch racing `clauth delete`
+        // — must bounce off cleanly instead of stranding the machine
+        // half-switched with the live link already destroyed.
+        if config.find(name).is_none() {
+            bail!("profile '{name}' not found");
+        }
         if config.is_active(name) {
             return Ok(());
         }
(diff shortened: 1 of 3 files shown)


uwuclxdy:
The gap this closes is real, thanks! The guard placement has a hole tho: it covers 1 of the 3 switch primitives. `switch_profile_discard` and `switch_profile_reconciled` have no existence check, both reachable from the exact surfaces the PR body names (the MCP `switch` tool, a CLI switch) whenever the outgoing active is diverged. On a ghost target `force_link_profile_credentials` tears down the live `.credentials.json` before `finish_switch` notices, the discard path takes no prior snapshot, so the uncaptured re-login is gone for good. Strictly worse than the bug being fixed.

The new test only drives the guarded branch (a `LinkedTo` live file), so the suite stays green with the diverged hole live. A diverged-active + ghost-target test would be red on this branch today.

I'll hoist the existence check into `switch_profile_cli` + `switch_profile_noninteractive` ahead of the diverged classify so all three primitives get covered in one place, plus that regression test. Pushing to this branch directly, done with claude.

PR #37, opened 2026-07-12
fix(scheduler): active profile's 429 back-off caps at 2x cadence

xingfanxia:
## The problem

The 429 back-off ladder treats every profile the same: `interval + 10s × 3^(streak-1)`, capped at 15 minutes. That's right for **idle** profiles — a sustained per-account rate limit needs a growing back-off to drain, and nobody is looking at those rows anyway.

It's wrong for the **active** profile. Its account's rate-limit window is dominated by the running claude's own API traffic — clauth's one poll per slot neither pins nor drains the limit. A deep slot buys nothing except staleness on exactly the row the user is watching, and exactly the numbers the auto-switch decision reads.

Observed live (2026-07-12): the endpoint had recovered (200s on direct probes) while the active account sat out a 14-minute slot still displayed as `RateLimited` — and auto-switch logic kept reasoning from that stale state.

## The fix

In `next_slot_deferral`, the active profile's ladder caps at one extra interval (2× cadence):

```rust
if active {
    ladder = ladder.min(interval_ms.saturating_mul(2));
}
```

- Idle profiles keep the full drain ladder, unchanged.
- A real server `retry-after` still wins over the cap (`hint.max(ladder)` is untouched).
- The flag is read at apply time under the config lock, so a switch mid-flight lands the right cadence on the right profile.

## Verification

Three new tests pin the contract: the cap engages on deep streaks for the active profile, a real server hint still overrides it, and shallow streaks (within the cap) behave identically for active and idle. Full suite: 729 passed, clippy clean.

Note: independent of #36; both touch different files. If you'd rather see this soak longer, we've been running it in production on our fork since 2026-07-12 — it's what un-stuck a permanently-`RateLimited` active row.

### DIFF
--- a/src/usage/scheduler.rs
+++ b/src/usage/scheduler.rs
@@ -837,15 +837,26 @@ fn rate_limit_backoff_ms(streak: u32) -> u64 {
 /// (observed 2026-07-11: hours of uninterrupted per-account 429s that only a
 /// growing back-off can drain). Capped at [`MAX_RETRY_AFTER_MS`]. Non-429
 /// outcomes: no defer.
+///
+/// The ACTIVE profile's ladder caps at one extra interval (2× cadence): its
+/// account's rate-limit window is dominated by the running claude's own API
+/// traffic, so clauth's one poll per slot neither pins nor drains it — a deep
+/// slot only buys staleness on the exact row the user is watching (observed
+/// 2026-07-12: the endpoint recovered while the active account sat out a
+/// 14-minute slot as `RateLimited`). A REAL server `retry-after` still wins.
 fn next_slot_deferral(
     rate_limited: bool,
     retry_after: Option<Duration>,
     streak: u32,
     interval_ms: u64,
+    active: bool,
 ) -> IntervalMs {
     let hint = retry_after.map(|ra| ra.as_millis() as u64);
     let target_ms = if rate_limited {
-        let ladder = interval_ms.saturating_add(rate_limit_backoff_ms(streak));
+        let mut ladder = interval_ms.saturating_add(rate_limit_backoff_ms(streak));
+        if active {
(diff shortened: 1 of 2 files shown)


uwuclxdy:
The staleness fix is real (the 14-min stale `RateLimited` row you captured), keeping it. The justification needs a correction tho. It changes the shape of the fix.

`FetchStatus::RateLimited` has exactly one producer: a 429 on the `/usage` poll itself (`fetch.rs`, the sole `FetchError::RateLimited` origin). An account at 100% on its 5h/7d window still gets a 200 with data. So the window behind a red row is the `/usage` endpoint's own per-account request throttle, which clauth's polls are the sole filler of; the running claude's `/v1/messages` traffic never touches it. #30 (merged earlier today) measured that rejected polls count toward that window: ~50 min pinned while polling at cadence, 4 min to drain once the ladder spaced retries out. The retry-after valve gives no cover here either: `/usage` answers every 429 with `retry-after: 0`, so `hint.max(ladder)` degenerates to the capped ladder. During a genuine storm the cap alone sets active re-poll spacing at ~3 min, which sits exactly in the gap #30's data didn't measure (pinned at ~45 s spacing, drained at minutes-to-15-min).

So: merging with a guard on top. The cap only engages while the streak is shallow (`streak <= 6`); a sustained storm releases to the full drain ladder. That still covers the fast-recovery case your incident was, while bounding the rejected-poll budget fed to the request window when the endpoint is genuinely storming. Docs reconciliation (`internals.md` deferral spec + a learning) rides along. Pushed to this branch, done with claude.

Principled follow-up, out of scope here: keep the poll cadence untouched, mark a deep-slot active `RateLimited` as stale in the TUI + have `scan_auto_switch` distrust it. Kills the stated harm with zero extra `/usage` load. lmk if you wanna take that one.

PR #38, opened 2026-07-12
feat(fallback): weekly (7d) window gates the auto-switch at a configurable soft line (default 98%)

xingfanxia:
## The gap

The weekly (7d) window previously gated only at the 100% hard cap. In practice that's too late for the auto-switch chain: an active account at **7d 98% / 5h 40%** sat unswitched — the 5h trigger never fired, the 7d window kept climbing, and the moment it topped out the account was bricked for **days**, not hours, mid-session. (Observed live: an account hit exactly 98.0% → 100% within one session.)

## The change — two commits

**1. `feat(fallback): weekly window gates at a 98% soft line`**

The line becomes a soft 98% applied to **both directions** of the walk:

- *Active trigger*: either window past its line fires the hop — 5h ≥ per-member threshold **or** 7d ≥ weekly line — so the switch lands while there's still weekly room to land it.
- *Candidate acceptance*: the walk never picks a member past the weekly line (it would re-trigger next tick). Last-resort members still accept, as the chain's parking spot.
- *Recovery + resume captions* key on the same line, so a weekly-dead member isn't "recovered" by a 5h rollover.

Lapsed 7d resets still read as renewed quota; per-model `weekly_scoped` windows still never gate.

**2. `feat(fallback): weekly exhaustion line is configurable (50-100, default 98)`**

The 98 graduates from a const to `AppState.weekly_switch_threshold` — **chain-global** by design (the line protects the *chain's* ability to land a hop, not per-member taste), absent = 98, valid band 50–100. The accessor clamps an out-of-band hand-edit in `profiles.toml` back to the default rather than silently disabling the weekly gate.

Config tab grows a `weekly limit` row with the same segmented-control grammar as the refresh row: space steps the 90 / 95 / 98 / 100 presets, ⏎ opens an inline custom-value editor (decimals allowed, range-validated, DANGER treatment on invalid input). Every fallback predicate takes the line as a parameter; `ChainSnapshot` carries it so the scheduler reads one consistent value per walk.

## Verification

- `weekly_line_is_configurable_chain_wide`: line at 90 excludes a 7d-92% member from the walk; default accepts it; a garbage 120.0 in state clamps back to 98.
- Both-direction tests: active trigger fires on 7d alone; candidate past the line is rejected; lapsed reset renews.
- Full suite: 729 passed, clippy clean.

Running in production on our fork since 2026-07-11 — same-day it correctly weekly-excluded an account that landed on exactly 98.0%.

Independent of #36/#37 file-wise except `src/usage/scheduler.rs`, which #37 also touches — whichever merges second rebases trivially (disjoint hunks).


uwuclxdy:
Correctness here is clean, thanks! The boundary routes uniformly through `weekly_blocked_info` for trigger / candidate / recovery alike, serde back-compat matches the `default_divergence` precedent exactly, units stay percent end to end, every changed call site is updated. Tests pin the 98.0 boundary from both sides.

One semantics change I want walked back: the soft line gates the wrap-off `Off` decision too. With wrap-off on and the whole chain at >=98% weekly (fresh 5h windows), clauth signs out running claudes ~2% of weekly before the API would refuse. On the switch path the early line buys safety headroom; on the halt-everything path it buys nothing, you just lose the tail. So the split I'm going with: soft line keeps gating the switch trigger + candidate walk + recovery pick, wrap-off `Off` keys on real exhaustion (100% / wall-clock) as before.

Smaller bits I'll fold in: a wrap-off-under-soft-line test, `parse_weekly_pct` boundary pins, the preset ladder deduped into one shared const, the "clamp" wording in `profile.rs` (it resets to default rather than clamping). Pushing to this branch directly, done with claude.

PR #40, opened 2026-07-13
feat(scheduler): distrust a deep-slot stuck RateLimited active (the #37 follow-up)

xingfanxia:
Taking you up on the principled follow-up you offered when merging #37:

> keep the poll cadence untouched, mark a deep-slot active `RateLimited` as stale in the TUI + have `scan_auto_switch` distrust it. Kills the stated harm with zero extra `/usage` load.

This is the consumer-side complement to #37's poll-side cap.

## The wedge

`scan_auto_switch` only trusts a `Fresh` active read to decide a switch (via `decision_fresh`). A stuck-`RateLimited` active — one whose `/usage` polls stay 429'd past the active cap and can therefore never return `Fresh` — makes the scan bail every tick, so the daemon **wedges on the rate-limited account and never rotates away**. It's the `RateLimited` analogue of the auth-broken wedge that already has an `active_broken` bypass.

## The fix — two parts, mirroring the auth-broken bypass

1. **Decision.** New `is_stuck_rate_limited(status, streak)` = `RateLimited && streak > ACTIVE_CAP_MAX_STREAK` (reusing #37's own cap boundary as the deep-slot line — within the cap window retries are kept frequent, so only *past* it is a still-`RateLimited` read genuinely stuck). `scan_auto_switch` gets a third bypass beside `active_broken`. Crucially the bypass only lets the walk **run** — unlike auth-broken it still faces `next_auto_switch_target`'s last-known exhaustion gate (guarded by `five_hour_live` + the weekly reset). So a genuinely-spent stuck active rotates away (wedge fixed) while a throttle artifact with real headroom **stays put** (no false switch). `fallback.rs` is untouched — that gate is the safety.

2. **Display.** The same predicate publishes an additive per-profile `status.json` `stale: bool` (schema stays 1; always `false` for the single-shot `status --json`), threaded via `LiveSignals.streaks`, so a menu-bar reader can dim a distrusted reading instead of rendering the frozen number as truth. One shared predicate → decision and display can't drift. Documented in `wiki/daemon.md`.

Zero extra `/usage` load — pure reinterpretation of data the scheduler already holds.

## Tests

`scan_auto_switch` walks off a deep-slot stuck-**spent** active; stays on a throttle artifact (LIVE headroom) **and** on a lapsed-stale-high window — the real post-storm shape, and the exact point this diverges from the auth-broken bypass, which *walks* on that same store; still waits for `Fresh` on a shallow streak. `build_status` `stale` is per-profile (proved with two profiles in one body), additive, and false for the single-shot. Full suite + `fmt` + `clippy` clean.

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -418,6 +418,14 @@ impl Daemon {
             .lock()
             .map(|m| m.clone())
             .unwrap_or_default();
+        // Snapshot the 429 streaks so build_status can publish `stale` (a
+        // deep-slot stuck RateLimited). Lower rank than config, like the stores
+        // above — snapshot + release before the config lock.
+        let streaks_snap: HashMap<String, u32> = self
+            .rate_limit_streaks
+            .lock()
+            .map(|m| m.clone())
+            .unwrap_or_default();
         // The in-flight switch target (accepted, not yet applied), so the UI
         // shows in-flight truth instead of a timing heuristic. Snapshot + release
         // before the config lock, like the freshness stores above. The set holds
@@ -431,6 +439,7 @@ impl Daemon {
         let live = LiveSignals {
             status: &status_snap,
             next_refresh: &next_snap,
+            streaks: &streaks_snap,
             pending_switch: pending_snap.as_deref(),
         };
         let body = {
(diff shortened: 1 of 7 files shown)


uwuclxdy:
Merged, thanks! The lapsed-window test is exactly the case I wanted pinned (a false switch there logs out every running claude over a reset account). Predicate boundary checks out against #37's cap release too.

Took the TUI half of the follow-up myself: the rate-limited status line now counts the retry (`· 7th retry in 840s`), so a deep slot reads as stuck from the number alone (67ee5ac, claude again).

PR #46, opened 2026-07-16
fix(daemon): don't defer switches behind Claude Code's logged-out shell

xingfanxia:
Caught this live on 2026-07-15, on a headless box running the daemon.

## The wedge

When Claude Code's **own** token refresh dies, it doesn't delete the live `~/.claude/.credentials.json` — it writes a **logged-out shell**: `accessToken: ""`, `refreshToken: ""`, `expiresAt: 0`, with unrelated keys like `mcpOAuth` kept in place. That shell still classifies `LinkState::Diverged`, so every guard built on "diverged and unsaved" treated an empty file as a login someone must resolve:

- the daemon deferred every queued auto-switch (`active '<name>' has unsaved credentials; resolve in the TUI`) — on a headless box, forever;
- `switch_profile`'s non-force link refused to replace the file, so even the escape hatch was wedged;
- the TUI banner nagged about credentials that don't exist, and a configured `default_divergence: Overwrite` would happily "capture" the blank tokens **over the profile's stored login**, destroying it.

Meanwhile every running `claude` sat at "Login expired" behind a file with nothing in it.

## The fix

The divergence guards exist to protect a live login the operator may hold nowhere else. A shell holds no login, so it forfeits that protection — and nothing else changes.

- `claude::live_login_is_empty(&creds)`: no OAuth block, or access AND refresh both absent/blank.
- `claude::live_credentials_are_shell()`: the live file **parses** to such an empty login.
- Every "unsaved credentials" predicate gets the one-line exemption: the daemon deferral (`active_diverged_unsaved`), its TUI twin, `switch_profile`'s uncaptured-relogin guard (so the switch takes the force-link and replaces the shell wholesale), the CLI `[Y/n]` reconcile prompt (capturing blank tokens over the outgoing store would destroy it), and the headless/MCP divergence-default branch. The TUI's 1Hz poll also skips the shell: no banner, and `default_divergence` can never capture it.

**Mid-write caution retained**: only a file that *parses* to an empty login is a shell. An unreadable or torn live file is "possibly a login mid-write" and keeps the full deferral, exactly like a real diverged login. A real re-login (either token non-blank) keeps every existing protection — the reclaim-vs-protect line is "the endpoint of both tokens is provably nothing", decided offline by the file's own content.

## Tests

- `live_login_is_empty` truth table (blank pair / blank access only / no OAuth block → empty; any single live token → not empty).
- `live_credentials_are_shell` requires a *parsed* empty login: missing file, torn JSON, and a real login all read false.
- Daemon: a queued switch **proceeds** over a shell (lands the target's stored login in the live slot, never captures the blanks over the outgoing store, leaves nothing queued); a torn live file **still defers** and stays queued; the existing `drain_pending_switch_skips_on_active_divergence` keeps pinning the real-divergence deferral.
- TUI: the poll ignores a shell — no banner, and `default_divergence: Overwrite` leaves the stored chain untouched.

944 tests, `fmt`, `clippy -D warnings` clean.

This is the minimal standalone piece of a bigger incident class (a diverged live login whose owner can't be proven locally); I'll write that up separately as an issue rather than smuggle it in here.

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -84,10 +84,8 @@ pub(crate) fn switch_profile(config: &mut AppConfig, name: &str) -> Result<()> {
         // switch. (Interactive callers already route a real divergence to the
         // reconcile path, so this branch is only reachable uncaptured via the
         // scheduler — where refusing, not dropping, is the safe outcome.)
-        // Claude Code's logged-out shell (both tokens blanked) is NOT an
-        // uncaptured re-login: there is no login in it to strand, and the
-        // non-force guard would otherwise refuse to replace the empty file,
-        // wedging every switch until a human resolves nothing in the TUI.
+        // A logged-out shell holds no login to strand, so it forfeits the
+        // refuse-guard (which would otherwise wedge the switch on an empty file).
         let uncaptured_relogin = match config.state.active_profile.as_deref() {
             Some(active) => {
                 matches!(classify_credentials_link(active)?, LinkState::Diverged)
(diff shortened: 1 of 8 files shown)


uwuclxdy:
merged. added two follow-ups on top:

- trimmed the shell-exemption comments. the `both tokens blanked` explainer was repeated at all 5 call sites, so i cut them to one-liners and left the canonical version on `live_credentials_are_shell` / `live_login_is_empty`.
- pinned the MCP `switch` `Overwrite`-over-shell branch with a test (`divergence_over_a_shell_switches_without_capturing`). the writeup listed the headless/MCP branch as fixed but only the daemon + TUI paths had coverage.

thanks for the detailed writeup, made triage easy. both commits are claude 🥺 lmk if i trimmed those comments too hard

PR #49, opened 2026-07-16
feat(tui): narrow-terminal (phone ≈45 col) adaptation — stack, don't truncate

xingfanxia:
Closes #48.

One breakpoint (`panes::NARROW_BODY_W = 60`, measured against the body width each frame) gates every change — **at or above it every rendering is byte-identical to today**, and the tests pin both halves of that contract (content survives at 45 cols; the desktop keeps its side-by-side shape at ≥60).

> **Disclosure** (same as #45): implemented with an LLM (Claude) driving under my direction; every frame below was rendered through the existing `TestBackend` harness and read by eye at 45×38 before/after, and the full suite + clippy + fmt ran green on this branch against `mommy`.

## What changes at < 60 cols

- **`panes::master_detail`** — new shared split for the five master-detail tabs (usage / setup / fallback / status / plugin) and the tokens Models view: horizontal `selector | detail` on desktop, **stacked selector-above-detail** at narrow (selector sized to its item count, capped at 40% of the body). At 45 cols the old split left the detail pane ~13 usable cells; stacked, both panes get full-width lines.
- **Tokens dashboard** — the side-by-side card pairs stack full-width (the `TODAY` card alone needs ~20 content cells; pairs at 45 cols crushed both).
- **Overview fallback-chain rows** — the gauge is sized from what the line actually has left (exact span budget, including 3-digit thresholds and 10+-member indexes) instead of a fixed 12 cells, so the trailing `42 / 80%` no longer hard-wraps onto an orphan line.
- **Footer** — measured degradation: while the row overflows, drop the rightmost non-essential hint; `←→ tabs`, `? help`, `q`/`esc`, and armed confirms always survive. (This also fixes a pre-existing clip at plain 80-col terminals, where the tokens tab's 86-col hint row lost `? help q quit` from the right edge.)
- **Header row 1** — the ` · ` separator now renders only when the gauge itself survives `gauge_fit` (budget math unchanged — the separator width is subtracted exactly as before, so fit decisions are bit-identical); previously a hidden gauge left a dangling `·`.
- **Status detail** — the right-aligned duration drops to its own line when there's no room to right-align, instead of gluing onto the age (`…2026-06-06ongoing`).
[...]

Issue #45, opened 2026-07-15
Enhancement: What about Codex Account Switching?


xingfanxia:
### Problem / motivation

Is this feasible? I've done some initial research it appears doable.
GPT 5.6 sol is a beast, loving it

### Proposed behavior

n/a

### clauth version

0.11.0

### Surface

TUI

uwuclxdy:
good idea, however i dont have any codex subscription so i couldn't test this or find it useful. it would also require implementing everything for openai endpoints and other stuff.. im not against the idea tho, i just feel like its a bit out of scope for now.

if we end up doing it tho, would appreciate a rough system design for it first here in this issue. if you're gonna use an llm to help you, could you also prepend your prompt or inital reasoning please?

and be careful with 5.6 as it tends to like rmrfing the home dirs 😹

uwuclxdy:
ref #44 when implemented, claude code will be able to use clauth mcp to delegate tasks to codex

xingfanxia:
> **Disclosure, as requested**: this design was worked out and implemented with an LLM (Claude) driving under my direction. The initial prompt was essentially: *"re-evaluate feasibility of codex account switching with claude-style seamless in-session fallback; if viable, design and build it incrementally with the same safety invariants as the claude side (loss-free switches, single-consumer refresh chains, no backend polling)."* Everything below was then verified against openai/codex source at `cbc83d9` / codex-cli 0.144.4 and shipped behind ~70 tests. And no home dirs were rm-rf'd 😹

Totally fair that this is out of scope upstream — I've built it in my fork ([xingfanxia/clauth](https://github.com/xingfanxia/clauth)), and here's the rough system design you asked for, in case it's ever useful.

## Design in one paragraph

[...]

PR #51, opened 2026-07-17
feat(codex): Codex multi-account support (fork's codex engine + enhancements, on 0.12.0)

xingfanxia:
Contributes the fork's **Codex (OpenAI) multi-account support**, rebased onto your `mommy` **0.12.0**. This is the code behind the design in #45.

> **Disclosure** (same as #45 / #49): implemented with an LLM (Claude) under my direction. The full suite + clippy + fmt run green on this branch against `mommy` (**1254 tests, 0 failures**), and the codex path was verified end-to-end against the live ChatGPT backend with a real paid account (below).

## Shape of the branch

The fork's history has been collapsed onto 0.12.0 into **8 subsystem commits** (from ~192 sprawling ones — squashed for reviewability; the tree is byte-identical to a straight merge, verified green). The marquee feature is Codex support:

- **`feat(codex)`** — the codex-exclusive engine: `src/codex/` (auth lens, OAuth refresh, usage/window routing) + `src/proxy/` (the localhost SSE injection proxy) + `docs/codex-support/` + a sandbox harness. **Cherry-pick this commit for the pure engine.**
- **`feat(core)`** — the harness axis (`Profile { harness: claude | codex }`), per-harness active slots, isolated `CODEX_HOME` starts.
- **`feat(scheduler)`** — per-harness pending-switch independence + the codex fallback chain.
- **`feat(daemon)`** — passive JSONL usage leg + `status.json` codex fields + proxy wiring.
- **`feat(tui)`** — codex rungs in the overview/chain/config panes. This also carries the narrow-terminal work from **#49** (they overlap in the render files), so merging this supersedes #49.

### The codex model in one paragraph
A `harness` axis on `Profile` splits accounts into claude/codex with independent active slots and fallback chains. A codex switch is a local, zero-network atomic rewrite of `~/.codex/auth.json`. Usage is read passively from the session-rollout JSONLs — no `wham/usage` polling. For true in-session fallback, `clauth proxy` is an opt-in loopback proxy that injects `Authorization` + `ChatGPT-Account-ID` per request and rotates+replays on a 429 before codex sees a byte; `clauth proxy --print-config` prints the config block (clauth never edits the live config).

## Verified against the real backend
The proxy was run end-to-end against live `chatgpt.com/backend-api/codex` with a real paid account, fully isolated on a throwaway host: real `codex exec` → `clauth proxy` (identity injected) → real backend → real model response. **OpenAI honors the proxy-injected identity.** One fix from that run is included — the proxy now forwards codex's `GET /models` refresh instead of 405'ing it (codex 0.144.5 degraded gracefully, but a stricter client could break).

## Caveats for review
- **Codex is untestable without a ChatGPT/Codex subscription** — the codex legs can't run on an account without one. The **claude-side behavior is byte-identical**, guarded by the full suite.
- This is a **complete, compilable, tested** branch, not a codex-*only* extraction — codex is woven into the harness infrastructure it needs, so a clean codex-only diff isn't separable from the shared files. Cherry-pick `feat(codex)` for the engine, or take the branch for working codex support.
- Everything already merged from this fork (keychain, OAuth login, scheduler hardening, shell-unwedge) is in your 0.12.0 and therefore **not** in this diff — this is only the codex (+ narrow-tui) delta.


uwuclxdy:
Will need some refactors before we can merge this. Also, it's not bad, but I'll still take another look and make some changes first, okay? I'll let you know when I'm ready to review this, don't worry about the conflicts until then

xingfanxia:
> Will need some refactors before we can merge this. Also, it's not bad, but I'll still take another look and make some changes first, okay? I'll let you know when I'm ready to review this, don't worry about the conflicts until then

sounds good

uwuclxdy:
would you agree to have a clauth mode switcher so maybe we can support other harnesses in the future more easily? what i mean is, have `c codex`/`c claude` in the hint bar (depending which harness is currently selected) and show something like `clauth • codex`/`clauth` in the header? this would allow to port all features (not just account switching and fallback) of clauth for other harnesses, starting with codex. what do you think? how should daemon handle this? am i missing anything?

xingfanxia:
Yes — enthusiastically. This branch is effectively running that design already, so I can answer from what held up and what bit us.

**The one structural call that matters: harness is a profile axis, not an app mode.** `Profile.harness: claude | codex`, and everything routes off it per profile. The `c codex`/`c claude` switcher and the `clauth • codex` header are exactly right — but as a *view filter* over which harness the tabs render, not a state the daemon is "in". The reason: both harnesses genuinely run at once (I have claude and codex sessions side by side all day), so a global mode would serialize two concurrent things. Concretely that means the state model splits per harness:

- **per-harness active slot** — `active_profile` for claude, a separate active for codex; a switch on one never touches the other
- **per-harness fallback chain + walk** — the codex chain reuses the walk engine wholesale, just over codex members
- **per-harness pending-switch queue** — one `VecDeque` keyed by harness, or a queue entry carrying it; a deferred codex switch must not block a claude switch

Once state splits that way, the daemon doesn't "handle" the mode at all — it iterates profiles and dispatches on the axis per subsystem. The dispatch points that actually fork:

1. **Switch mechanics** — claude = keychain/live-file link; codex = `~/.codex/auth.json` write, and it's *session-boundary*: a running codex keeps its in-memory token, so the switch lands on next start. Worth surfacing that in the TUI so a switch doesn't look like a no-op. (For in-flight swaps we ended up with a localhost injection proxy that rewrites the bearer per request and rotates-and-replays on 429 — works well, but it's a separate feature, not something the switcher needs.)
2. **Refresh discipline** — this is the sharp edge. Codex refresh tokens are *single-use rotating*: any second writer (the codex CLI itself refreshing while the daemon does) kills the chain permanently, not transiently. We run a machine-wide flock single-writer per chain and only refresh parked profiles (the live one belongs to the CLI). Claude-side habits about "re-mint on demand" do not transfer.
3. **Usage** — no `/usage` endpoint; parked-account usage comes from `GET chatgpt.com/backend-api/wham/usage` with the stored access token (codex CLI itself polls it ~60s, openai/codex#10869). Windows are duration-keyed (`limit_window_seconds`) rather than named, and the live plan tier rides the same response — the stored id_token's plan claim goes stale on upgrades.
4. **Feature gating, not porting** — some claude features must *skip* codex profiles rather than be ported: settings sync and the session machinery are meaningless against `CODEX_HOME`, and letting them touch codex profiles corrupts things. So the axis needs to be visible to every feature, even ones that only support one harness — "is this profile mine to touch" as a first-class question.

The "am I missing anything" item I'd flag: auto-start parity. Starting a *parked* codex account (without switching the live auth.json) needs an isolated `CODEX_HOME` per profile plus a lease/adopt-back dance so the account's usage and refreshes stay attributed. That's the piece that took the most iteration on my side.

Happy to upstream this incrementally once #55/#58 land — the natural first slice is the harness axis + per-harness active/chain state (mechanical, no codex-specific code yet), then switching, then refresh/usage. The fork has all of it running if you want to poke at the shape: harness axis, standby refresh, chain walk over codex members, JSONL + poll usage legs, proxy.

uwuclxdy:
couldn't `CODEX_HOME` be repointed to a `~/.clauth/profiles/{}` like claude code is? we could spawn a session like that for things that we need to per profile and also `clauth start {codex account}`. i haven't tried codex but i plan to since they give some free access and i'd also like to know if you'd be okay with rewriting this branch on top of v0.14 release when it lands and if you could help with reviewing changes i'll make to codex mode when we get there. i think it is possible to achieve feature parity with codex, depending on how differently they do things (maybe some things will stay claude code specific if codex already does it by default). also, i'd like to have profiles suffixed by the account type, eg. `~/.clauth/profiles/{name}-cc` for claude code and `~/.clauth/profiles/{name}-c` for codex. the agent sees these paths in its system prompt so i would be against suffixing them with something longer. let me know what you think before starting

xingfanxia:
Yes to all three — and the `CODEX_HOME` question is another one this branch already answers, so here's what a few weeks of running it says.

**`CODEX_HOME` repointing: it works, and it's exactly how codex starts run here.** One nuance vs. the claude shape: we pin it to a subdir, `~/.clauth/profiles/{name}/codex-home/`, not the profile dir itself — the profile dir also holds the store files and lease dirs, and codex creates its own tree inside its home. The isolated home is three entries: `auth.json` (the profile's chain, seeded from the store at acquire), `config.toml` (a **copy** of the operator's, not a symlink — codex writes through it, and a symlink would mutate the real `~/.codex/config.toml`), and `sessions/` (codex-created, isolated history). `clauth start {codex account}` then is just: acquire lease → seed home → spawn `codex` with `CODEX_HOME` pinned. The sharp edges we hit, so you don't have to:

- codex canonicalizes `CODEX_HOME` and hard-errors on a missing dir — create it before spawn, and canonicalize the path so a symlinked `~/.clauth` works.
- scrub any inherited `CODEX_HOME` from the env before pinning yours, or a nested start leaks the parent profile's home into a different profile's session.
- check `cli_auth_credentials_store` first: in keyring mode codex ignores the file in the isolated home entirely — we refuse unless it's the default `file` mode.
- **the one that actually matters: never let two carriers hold one chain.** Codex refresh tokens are single-use rotating, so if the profile's account is ALSO the live `~/.codex` login, an isolated copy forks the chain and the next refresh on either side kills it permanently (`refresh_token_reused`). We refuse the isolated start when the account is live, and conversely capture/switch refuse profiles with a live session lease (flock per profile, plus a rotation guard across the acquire stamp so a standby refresh can't spend the chain mid-acquire).
- adopt-back: while a session runs, the isolated home IS the chain's carrier — one-way sync `codex-home/auth.json` → store on a 60s watchdog plus a final sync on drop; everything else (standby refresh, switch installers, chain walk) treats the leased profile as hands-off.

**v0.14 rebase: agreed, with one mechanical note.** This PR's head is my fork's main (which tracks your releases by true merges), so rather than rewriting it in place I'd cut a dedicated branch from v0.14 and port incrementally as a reviewable series — roughly: harness axis on `Profile` + per-harness state → codex store/capture → switch + isolated start → refresh/usage. That's also the shape that makes your codex-mode changes easy to review against, and yes — happy to review them as you go. Agreed on parity being reachable; the deltas that stay codex-specific in practice are the session-boundary switch semantics and the single-writer refresh discipline, both of which the design above absorbs.

**On `{name}-cc` / `{name}-c`: fine with the idea, two adjustments I'd push for.** First, keep a typed field authoritative: profiles here carry `harness = "codex"` in their per-profile config (absent = claude, immutable — cross-harness conversion is delete + recreate), and the dir suffix should be a naming convention applied at create time, not a parsed discriminator — otherwise a user profile named `work-c` becomes load-bearing ambiguity. Second, consider suffixing only the codex side and leaving claude dirs bare: every existing deployment has unsuffixed claude dirs, so `-cc` means a rename migration plus invalidating whatever paths agents have already absorbed into their prompts, while suffixing only the new axis is back-compat for free and matches "absent = claude". If you do want both suffixed, I'd spend one extra character on `-cx` for codex — `-c` next to `-cc` is one keystroke of visual difference, and "c" reads as claude at least as easily as codex.

uwuclxdy:
Design's settled, thanks for laying the shape out so completely. I verified the codex-specific bits against `codex 0.145.0` before locking it, and wrote the whole thing up as a spec at `docs/codex-plan.md` (in the repo). It's self-contained: the full design, the verified codex-0.145 behavior with source refs, the file and symbol anchors into the current tree, the pinned decisions, a reviewable-series breakdown, and the open questions. Read that first.

The ask: you own the implementation end to end, groundwork plus codex, on a branch cut from the `v0.14` tag (not out yet, latest is `v0.13.1`). I design and review. Ship it as the reviewable series in the plan (axis, then the two trait seams, then codex credential + isolated start, then refresh + adopt-back, then usage + fallback, then the parity surfaces), never one entangled diff. That's what made this PR hard to review the first time.

A few things in the plan are non-negotiable, since they're where a wrong move breaks claude for everyone:

- codex state lives in a SEPARATE `~/.clauth/codex-profiles.toml` that old binaries never read or write, so `profiles.toml` and every claude path stay byte-identical: no harness field in `AppState`, no migration, no dual-write. Harness is implied by which file a profile is in, not a discriminator parsed out of shared state. This is what dissolves the mixed-version corruption we chased through a few review rounds, an old binary can't touch state it can't see.
- Force `-c cli_auth_credentials_store="file"` on every codex spawn, not "refuse unless file". Under `auto`/`keyring` codex ignores the seeded `auth.json` and deletes it on the first refresh, so adopt-back reads a dead file and the chain dies on the next replay. Verified live.
- Single-writer refresh through the existing per-profile `RotationGuard`; `refresh_token_reused` is permanent, not transient.
- claude paths stay bit-identical, gated by the existing suite.

The open questions in the plan are yours to propose on: `delegate` via `codex exec`, whether a codex profile shares the operator's global codex config, the usage-window model mapping, whether auto-start even applies to codex, the proxy's scope, and `codex login` reuse. I'll answer or we settle them together.

Fork's the wire reference. Thoughts on this shape?

Feel free to add me on Discord or Signal for faster iteration https://uwuclxdy.dev/#contact

xingfanxia:
Taking it — groundwork plus codex, branch cut from `v0.14`, shipped as the six-part series. The design's yours and the pinned decisions stand as written; everything below is gaps and answers, not re-openings.

Two of the pinned decisions are better than what this branch runs, and I'd rather say so plainly since I'll be deleting my own code to adopt them:

- **Decision 5** (force `-c cli_auth_credentials_store="file"`) beats the fork's refuse-unless-file (`runtime.rs:247`). Refusing punishes a keyring user for a setting that only matters inside a home we control. One scope note in gap 2.
- **Decisions 1/3** (separate `codex-profiles.toml`). The fork carries `harness` on the per-profile `config.toml` plus `active_codex_profile`/`codex_fallback_chain` inside `AppState` (`profile.rs:495-507`). That only works because fork dirs are bare — the moment dirs are `-cx`, the dir path depends on the harness and the harness can no longer live in a file inside the dir. The file split is the right resolution, and it dissolves the mixed-version write-hole for free rather than patching it.

## Gaps

### 1. Age-scheduled standby refresh is blind to the actual death mode (phase 4)

This is the one that matters. Phase 4 as written inherits the claude shape: refresh parked chains on an age schedule. We ran exactly that for weeks, then did a root-cause study on 2026-07-23 after a parked codex account went permanently dead.

The finding: **TTL never kills a codex chain — a second holder does.** Because the refresh token is single-use rotating and reuse is permanent, a parked chain dies the moment anything outside clauth rotates it (an interactive `codex` run against the same account, another machine, a stale isolated home that never adopted back). The chain is dead *immediately*, but an age-scheduled refresher doesn't look until its 48h/7d timer fires, so the account sits in the fallback chain advertising itself as healthy and gets selected. The signal was there the whole time — `wham/usage` had been returning 401 for that profile — the refresher just wasn't listening to it.

Recommendation: phase 5 wires `PollError::Unauthorized` into a kick set that phase 4's standby leg drains on its next tick and force-refreshes, bypassing **only** the age gate — the `RotationGuard`, candidate-set membership, and the apply-time chain-identity re-check all still apply, so it stays inside decision 6's single-writer discipline. It needs a consecutive-kick breaker (we use 2, reset on any successful poll) or a wham-side auth quirk on a healthy chain rotates it forever; a genuinely dead chain exits through `Permanent` → quarantine instead. Fork implementation is in `usage/scheduler.rs` (`codex_auth_kicks`, `CODEX_KICK_STREAK_MAX`) if it's useful as a reference.

Name collision worth heading off: this kick is a health-triggered refresh. Q4's kick opens a claude 5h window. Unrelated mechanisms.

### 2. Decision 5 covers spawn, not capture

Forcing the flag fixes every clauth-controlled spawn. It does not fix the *capture* path — `clauth login <p> --codex` reads the operator's real `~/.codex/auth.json`, and under `keyring`/`auto` that file is absent or stale by design, so capture silently snapshots nothing (or a dead chain) for exactly the keyring user decision 5 set out not to block. Capture needs its own arm: read the keyring, or refuse with a message naming the setting. The fork refuses at `actions.rs:1386` — that one stays, it's the *spawn*-side refusal that should go.

### 3. The `enforce_clauth_perms` exemption needs its other half

Confirmed on our side: the fork has the identical unguarded recursive walk (`profile.rs:1237`), reached from `load_config()`, so it runs on every entry point same as yours. Two constraints on the fix so the exemption doesn't loosen what it's protecting: skip *descending into* `codex-home/` rather than skipping the dir node alone, and make `auth.json`'s writer responsible for 0600 (fork uses `atomic_write_600`) instead of relying on the sweep to retighten it.

### 4. Seed and adopt-back must be byte-preserving

`auth.json` should move as an opaque blob in both directions, never re-serialized from a parsed subset. `last_refresh` is the reason: codex self-refreshes when that stamp is older than the day-scale interval, so dropping or zeroing it makes a freshly seeded isolated codex refresh on startup and race the daemon's standby refresh for the same chain — the two-carrier death, self-inflicted at seed time. Same argument for `OPENAI_API_KEY` and anything codex adds later.

### 5. The mixed-version window is silent

Decision 3 is right that an old daemon simply ignores codex profiles. The user-visible result is "codex support does nothing, and nothing says why" — and #57's singleton makes it stickier, since the old daemon holds the lock and the new binary's `daemon --standby` exits cleanly by design. Cheap fix: publish the daemon's own version in `status.json` (additive, schema stays 1 — the fork already carries `clauth_version` there) and have a new CLI print one line when the running daemon predates codex support.

### 6. Minor: say `-cx` is reserved

Worth putting the reservation in `validate_profile_name`'s error text, or the rejection reads as a mysterious name rule.

## Open questions

**Q1 — delegate / `clauth start {codex}`.** `start` is interactive `codex`; that's shipped and stable in the fork. `delegate` is the one item I'd cut from the series: `codex exec`'s output contract isn't `claude -p`'s, so the MCP tool needs a per-harness result formatter, and it's the only piece here I have zero production evidence behind — the fork's `delegate` is claude-only (`mcp/mod.rs:105`). Ship it as its own PR after the series; until then the tool should error clearly on a codex profile rather than silently doing something claude-shaped.

**Q2 — shared config in the isolated home.** Allowlist, not bare and not a blanket "symlink the read-only bits". MCP servers come free with the `config.toml` copy — they're configured in it — so the real question is the asset dirs. A heavily-used `~/.codex` here holds `skills/`, `rules/`, `agents/`, `templates/`, `references/`, `AGENTS.md`: that's where the operator's actual behavior lives, and a bare home is a different assistant, not an isolated one. Everything else in that tree is writable per-session state — `sessions/`, `history.jsonl`, four sqlite DBs plus their WALs, `log/`, `cache/`, `tmp/`, `installation_id` — and must stay isolated; a symlinked `sessions/` defeats the isolation outright and a symlinked sqlite invites concurrent writers. So: symlink a pinned allowlist, copy `config.toml`, isolate the rest. I'd leave `hooks.json` out of the allowlist by default, since hooks execute code.

**Q3 — usage model mapping.** Reuse the named model and map by duration; no codex-specific window shape needed. The fork's `route_windows` (`codex/usage.rs:302`) classifies each limiter window on `window_minutes` — over 24h goes to the weekly slot, otherwise the 5h slot — with a positional fallback (primary → short, secondary → weekly) when the duration is absent, and a collision rule that pushes the second window to the other slot. `rate_limit_reached_type` is then renamed to whichever slot its window actually landed in so consumers read one vocabulary. One mapping serves both legs (passive JSONL and `wham/usage`), and the TUI renders the existing 5h/7d columns unchanged.

**Q4 — auto-start.** Claude-only. `wham/usage` reports a parked account without opening anything, so there's nothing for a kick to open, and codex has no equivalent window-opening endpoint.

**Q5 — plan-tier staleness.** Solved by the usage poll: the `wham/usage` body carries a top-level `plan_type` that is the live counterpart of the stale `chatgpt_plan_type` id_token claim (fork: `codex/poll.rs:64-68`). Treat the polled value as authoritative and the JWT claim as the fallback for an account not yet polled.

**Q6 — `clauth proxy`.** Defer, standalone, agreed. One lesson if it ever lands: a cached "spent" reading has to be an advisory rank in selection, never a synthetic 429 — a stale Plus-era figure on an account since upgraded to Pro wedges the chain shut.

**Q7 — codex login.** Reimplement PKCE; don't shell out. `codex login` writes the LIVE `~/.codex/auth.json`, so using it to populate a profile clobbers the operator's live login and forks the chain — decision 6's two-carrier death, triggered by the login path itself. The fork's `codex/login.rs` (~200 LOC, verified against openai/codex `9ff47868` / cli 0.144.5) mints a store snapshot without ever touching the live file. Traps worth pinning in the spec:

- the registered loopback redirect ports are fixed — 1455, fallback 1457, path `/auth/callback`. An ephemeral port doesn't work.
- the code exchange is **form-urlencoded**; the refresh at the same endpoint is **JSON**. Same URL, different encoding per grant.
- write `auth_mode: "chatgpt"` explicitly — codex's `resolved_mode()` infers ApiKey mode from a bare `OPENAI_API_KEY` when `auth_mode` is absent.
- `tokens.account_id` comes from the id_token's `chatgpt_account_id` claim, as codex's own login does.
- the RFC-8693 secondary exchange that mints `OPENAI_API_KEY` is best-effort; its failure must not fail the login (codex uses `.ok()` there).

**Q8 — cross-harness CLI.** Agreed on disambiguating by file membership and erroring when a name lives in both. Two additions: the error should name both harnesses rather than just saying "ambiguous", and there should be an explicit override (`--codex` / `--claude`) so a script isn't stuck with a name it can't address.

**Q9 — Tokens tab.** Exclude codex in this series. `token_ledger` reads claude-shaped session JSONL; codex's is a different shape, and a half-wired tab that silently under-counts is worse than an honest exclusion. A codex cost lens is its own feature.

## Next

Blocked on the `v0.14` tag. Meanwhile I'll fix gap 3 fork-side, since it's the same bug, and I'll re-verify the phase-1 anchors against `mommy` HEAD when I cut the branch rather than trusting the line numbers in the plan.

uwuclxdy:
Rewrote `docs/codex-plan.md` against `mommy` HEAD. Read that first, the 2026-07-24 revision is stale in ways that matter: the multi-session fallback refactor landed since (per-session runtime trees, a live-session registry, a per-session credential swap executor), so every anchor got re-derived and four decisions flipped.

Two of them flipped because I went and read `rust-v0.145.0` instead of trusting my own notes:

- **codex refreshes on the access token's clock, not on a day timer.** `should_refresh_proactively` (`manager.rs:2506`) tests the access_token JWT `exp` first and refreshes within `CHATGPT_ACCESS_TOKEN_REFRESH_WINDOW_MINUTES = 5`; the 8-day `TOKEN_REFRESH_INTERVAL` is the fallback for an exp that won't parse. My spec said day-scale, which was reading the fallback as the rule. A live codex session rotates the chain routinely.
- **codex already does read-before-spend.** `refresh_token()` (`manager.rs:2366`) reloads `auth.json` from disk and returns early with `Skipping token refresh because auth changed after guarded reload` when the on-disk auth moved under it. It POSTs only on `ReloadedNoChange`.

Put those next to `FileAuthStorage::save` writing in place, truncate + write, no rename (`storage.rs:202`), and the isolated-home design collapses into a symlink.

## What flipped

| was | is now | why |
|---|---|---|
| codex dirs suffixed `-cx` | bare, same as claude | the suffix only existed to let one name live on both harnesses |
| a claude `foo` and a codex `foo` may coexist | names unique across both files | `LiveTally` (`live_sessions.rs:129`), the pending-switch set (`scheduler.rs:163`) and all five per-profile cache files (`profile_cache.rs:38`) key on the bare name. One validation rule closes that whole class, and it removes the `--codex`/`--claude` override you asked for on `switch` |
| seed `auth.json`, adopt it back on a watchdog | `auth.json` is a symlink to `profiles/<name>/auth.json` | codex follows the symlink and re-reads before spending, so one physical file makes concurrent carriers safe. No seed, no adopt-back, no whitelist-diff of the home |
| day-scale self-refresh | 5-minute pre-expiry window | above |

Dropping `-cx` is the one I'd flag hardest, since it was my call originally. With unique names it bought nothing and cost a `-cx` resolution in every path derivation plus two reverse dir-to-name maps (`which.rs:145`, `runtime.rs:816`) that would have handed back `"<name>-cx"` as a profile name. Phase 1's path-layer work is now zero.

## Scope changes

- **MCP and the Plugin tab stay Claude-Code-only.** Both already fail closed: `load_config` builds from `profiles.toml` only (`profile.rs:1528`), so codex profiles are invisible to every MCP tool, and the Plugin tab has no per-profile selector to hand one to. Nothing behavioral, only copy: the init block's "Claude Code accounts", the plugin tab's "every profile's live sessions", the Tokens tab's missing qualifier, and `delegate`'s "profile not found" for a codex name that demonstrably exists. Your Q1 recommendation taken as written, `delegate` over `codex exec` is its own PR after the series.
- **Codex sessions join the live-session registry for accounting only.** Rows carry a harness tag, they count in the tally and gate delete/disable/rotation, they never enter `SessionSwap` or the per-session decision leg. A mid-session `auth.json` rewrite is a no-op the executor would publish as a successful member change. `--with-fallback` refuses on a codex profile.
- **Concurrent codex sessions allowed**, resting on the shared file rather than on any propagation code. `--isolated` symlinks `auth.json` too, so it isolates the config tree without forking the chain. That mirrors claude, where `canonical_credentials` (`runtime.rs:848`) already resolves both flavors to the same install source.
- Everything else you'd expect ports: chain, standby refresh, burn, additive published fields, the `c codex` filter.

## Your gaps

1. Taken, phase 4, including the 2-strike breaker and the naming note.
2. Taken. Your `actions.rs:1386` refusal stays; it's the spawn-side one that goes.
3. Taken, both halves: skip descending rather than skipping the node, and `auth.json`'s writer owns its own 0600.
4. Survives in reduced form. With no copy of `auth.json` anywhere there's nothing to seed or adopt, so what's left is that clauth's own writes preserve the keys it doesn't model (`OPENAI_API_KEY`, `agent_identity`, `personal_access_token`, `bedrock_api_key`). The in-place non-atomic write adds a second rule: a short or unparseable read is "retry", never "the chain is gone".
5. Taken, in phase 6.
6. Moot with bare dirs. The uniqueness rejection names which harness holds the name instead.

Q2 is yours with one change: `hooks.json` is opt-in per profile, default off, rather than out entirely. Q8 changes as above. Q3/Q4/Q5/Q6/Q7/Q9 stand as you wrote them.

Four claude-side latent defects fold into the series, since three of them only bite once codex exists: `is_exhausted_from_usage` reading an absent window as full headroom (`fallback.rs:951`), `per_profile_env_keys` readdir'ing the profiles root with no roster filter (`settings_sync.rs:180`), the rotate error path persisting nothing so the next tick replays the same token (`oauth.rs:740`, permanent death on codex rather than one `400`), and a stale `#[allow(dead_code)]` on `live_isolated_stores`.

## Refactors during the series

A `clean-rust` sweep found patterns this series would otherwise duplicate across two harnesses. They are NOT a gate and NOT your scope: they land on `mommy` while the series runs, sequenced before the phase whose diff each one shrinks. The phase map is in the spec under "Refactors landing in parallel." Nothing from that sweep reaches the PR thread.

Branch point: #59 lands first, then v0.14 gets cut, then you branch from the tag. It rewrites `ensure_installable` into a dispatch and adds a `session-token.json` sidecar, which is exactly the seam your phase 2 wraps, so branching ahead of it means writing that seam twice.

thoughts?

xingfanxia:
Read the 2026-07-27 revision. The four flips are all improvements and I'm not re-opening any of them — dropping `-cx` was clearly right once names are unique, and the symlink is the good kind of design change: it doesn't simplify the adopt-back subsystem, it deletes it. Reading `manager.rs:2366` before trusting your own notes is what made that possible, so thank you for going and doing it.

Three things below: the branch point (which v0.14 shipping has changed), one hazard the symlink introduces that I don't think the spec prices yet, and two smaller interactions.

## Branch point

You wrote: "#59 lands first, then v0.14 gets cut, then you branch from the tag." v0.14.0 shipped on 7-29 without #59, so that sequence no longer describes anything. Rather than guess, here's the state and a proposal.

#59 is **rebased onto `v0.14.0` and green** — 3 commits, `MERGEABLE`, `mergeStateStatus: CLEAN`, all four blockers and both design questions answered in the round-2 comment I just posted. So the seam your phase 2 wraps is settled code sitting on the tag, not a moving target.

Proposal: **I cut from `mommy` once #59 merges.** Your original reason still holds exactly — #59 rewrites `ensure_installable` into a dispatch and adds the `session-token.json` sidecar, and branching ahead of it means writing that seam twice. Nothing about that changed when the tag moved; only the ordering did.

If you'd rather not gate phase 1 on a review round, the fallback is: I branch from `v0.14.0` now and start phase 1, which touches none of the credential-install path — harness axis, the `codex-profiles.toml` state type, name uniqueness, `reload_fingerprint`, the `which.rs` dispatch, the `settings_sync` roster fix, the `live_isolated_stores` attribute, and the perms exemption. Phase 2 is where the seam lands, so it rebases onto whatever `ensure_installable` looks like by then. Either works for me; the first is cleaner and the second is faster. Your call — that's the one thing I actually need before starting.

(Gap 3 is already fixed fork-side, both halves: skip descending into the codex home rather than skipping the node, and `auth.json`'s writer owns its own 0600.)

## The symlink moves a non-atomic foreign writer onto the durable store

This is the one I'd want pinned in the spec before phase 3, because it's a new failure mode the seed/adopt-back design didn't have.

`FileAuthStorage::save` is truncate + write, in place, no rename (`storage.rs:202-218`) — you already have that in the verified-behavior section. What follows from combining it with the symlink is that **codex, a process clauth does not control, becomes a non-atomic writer of the profile's only copy of the chain.** Under seed/adopt-back the profile store was written by clauth alone, always through `atomic_write_600`; a crashed codex could only corrupt its own isolated copy.

The consequence is asymmetric with claude in a way that matters. A truncated `credentials.json` costs a re-login. A truncated codex `auth.json` costs the refresh token itself, and there is no second copy anywhere — no keyring (decision 6 forces `file`), no backup, no adopt-back source. Recovery is a browser re-login, which is the same cost as `refresh_token_reused` and which decision 7 treats as the thing worth designing around.

The spec's "a short or unparseable read is retry, never the chain is gone" rule handles a *reader* catching a partial write. It doesn't help if the writer dies mid-write, because then the file stays truncated and every subsequent read is that same "retry" forever — a wedge rather than a loss, but a permanent one.

I don't think this sinks the symlink; the concurrency argument for it is strong and the alternative reintroduces the two-carrier death. What I'd add is a cheap belt, and I'd rather propose it than implement it silently:

- clauth keeps a last-known-good copy of `auth.json` whenever it reads a well-formed one (it already has to read it for usage polling and refresh), written atomically, 0600, never symlinked.
- On a read that comes back short or unparseable **twice in a row across a guard acquisition** — i.e. not a live partial write — clauth restores from that copy under the `RotationGuard` rather than quarantining the profile.
- It is never a second *carrier*: it is never linked from any home and never written by codex, so it cannot fork the chain. It is only ever read to repair a file that is already unusable.

The window it doesn't close is a crash between codex's truncate and its write of a *newly rotated* pair, where the last-known-good holds the superseded refresh token and restoring it hits `refresh_token_reused`. That's unavoidable with an in-place writer and it fails the same way doing nothing does, so the belt is strictly better than nothing and never worse.

If you'd rather not carry the extra file, the honest alternative is to say in the spec that a codex crash during `auth.json` write is unrecoverable-by-design and the fix is `clauth login <p> --codex`. I'd just rather that be written down than discovered.

## Two interactions worth a line each

**1. Codex session rows and the live-session rotation refusal.** #59's first commit changes `rotation_blocked_for` from "any live `clauth start` session blocks rotation on macOS" to "any live session that LAUNCHED on something rotatable blocks it", reading a new `LiveSession.launch_store` field. It fails closed: a row without `launch_store` reads as rotatable and blocks.

Phase 3 registers codex sessions in the same registry. If those rows don't set `launch_store`, every live codex session will block rotation for its own profile on macOS — silently, and looking exactly like the age-gate stall gap 1 is about. The fix is one line at registration (`launch_store: Some(profiles/<name>/auth.json)`), and then the predicate answers correctly for codex too, since that file's content is what a codex session actually holds. Worth putting in phase 3's checklist rather than finding it in phase 4.

**2. The 5-minute stand-down and the health kick can deadlock a parked chain.** Decision 7 says clauth never refreshes a chain whose access token is inside codex's 5-minute pre-expiry window. That's right while a codex process is live for that profile — the race is lost by construction. For a **parked** profile there is no codex process, so nobody refreshes: clauth stands down, the access token expires, `wham/usage` 401s, and gap 1's kick fires a force-refresh that bypasses only the age gate — so it walks straight back into the stand-down, and the 2-strike breaker then trips on a chain that was healthy the whole time.

Two ways out, either fine: scope the stand-down to profiles with a live codex session (the liveness layer already answers that, and it's the only case the race exists in), or exempt the kick path from it explicitly. I'd take the first — it says the true thing — but the second is smaller.

## Otherwise

Everything else stands as written. Q2's `hooks.json` opt-in default-off is better than my "leave it out entirely". The four folded claude-side fixes are yours to sequence; I'll take them as their own commits as specified. And understood that the `clean-rust` sweep lands underneath me — I'll re-derive anchors at each phase rather than trusting line numbers, same as I did for #59's rebase.

Ready to start on your answer to the branch point.

uwuclxdy:
Start after #59 merges, i already reviewed the round-2 answers there. cut from `mommy`.

Belt is good: clauth keeps a last-known-good copy on every well-formed read (atomic, `0600`, never linked), restores after two consecutive bad reads under `RotationGuard`.

Both one-liners too: set `launch_store: Some(profiles/<name>/auth.json)` at codex registration in phase 3. scope the stand-down to profiles with a live codex session.

xingfanxia:
Status: phases 1–4 of the series are implemented and sealed locally (not pushed yet — I'll cut the reviewable series against current `mommy`). Before I do, an isolation sweep over the codex surface turned up six things that touch **your** pinned decisions rather than my implementation, so they're yours to rule on rather than mine to freelance. Four I've already fixed on my side because the fix was unambiguous; two I can't fix locally at all.

Everything below is against `rust-v0.145.0`.

## 1. `-c` is not codex's top config layer, so decision 6 does not hold as written

The spec pins "force `-c cli_auth_credentials_store="file"` on every clauth-controlled codex spawn", and my `codex_spawn_command` said in a comment that a later `-c` wins codex's layering. Both assume `-c` is authoritative. It isn't:

- `-c` becomes the `SessionFlags` layer, precedence **30** (`config/src/config_layer_source.rs`).
- `LegacyManagedConfigTomlFromFile` is **40**, `LegacyManagedConfigTomlFromMdm` is **50**, and `loader/mod.rs` pushes both into the same vec *after* SessionFlags, under the comment "as a config layer on top of everything else".
- `effective_config()` merges lowest-precedence-first, and `merge_toml_values` replaces scalars, so a managed value clobbers the `-c` value.
- The unix managed path is the machine-global `/etc/codex/managed_config.toml`, and `managed_config_default_path` explicitly discards `codex_home` — so redirecting `CODEX_HOME` per session does not isolate it.

Consequence: on an MDM-managed or root-writable Mac, `cli_auth_credentials_store = "keyring"` in that file silently defeats the pin. codex then ignores the symlinked `auth.json`, keyed by `sha256(canonical CODEX_HOME)` which is a fresh per-session home every launch, so it reports no credentials while `profiles/<p>/auth.json` goes stale — and clauth keeps rotating that stale file until it replays a spent single-use token into `refresh_token_reused`.

I can't fix this in clauth: the file is outside every boundary we control. What I'd propose is detect-and-warn — read the managed path at spawn and refuse (or loudly warn) when it sets either key we pin. Your call whether that's in scope for this series or a follow-up; I've left the code alone pending your answer.

## 2. A copied config key can *erase* the `-c` layer, so the forced-override idiom needed a belt

`debug.config_lockfile.load_path` makes codex rebuild the whole config from a single-layer stack — `ConfigLayerStack::new(vec![lock_layer], ..)` in `core/src/config/mod.rs` — and return. The `-c` layer isn't outranked, it's discarded. Since the session's `config.toml` is a verbatim copy of the operator's, an operator who set that key unpinned both forced overrides at once.

Fixed on my side by **sanitizing** the copy rather than only overriding it: `sqlite_home`, `cli_auth_credentials_store` and `debug.config_lockfile` are dropped from the session's config; an unparseable config still copies through verbatim. The forced `-c` stays as a second line. Flagging it because it changes the shape of decision 6 from "force it" to "strip it *and* force it", which I'd rather you saw stated than discovered.

(Residual, not fixed: a *trusted project's* `.codex/config.toml` can also set `load_path`, at precedence 25. Closing that means not propagating project trust into the session home, which is a real UX cost. I left trust propagating; say the word if you'd rather it didn't.)

## 3. Two env credentials outrank the linked `auth.json`

`load_auth` in `login/src/auth/manager.rs` resolves `CODEX_API_KEY` → ephemeral store → `CODEX_ACCESS_TOKEN` → *then* the persistent store, and says so itself: "API key via env var takes precedence over any other auth method". We were scrubbing `OPENAI_API_KEY` for exactly this reason and had missed its two siblings, so an operator with either exported had every `clauth start <p>` authenticate as something other than the profile it named, while clauth reported and burned that profile.

Fixed: both join the managed-env scrub, along with `CODEX_REFRESH_TOKEN_URL_OVERRIDE`, `CODEX_REVOKE_TOKEN_URL_OVERRIDE` and `CODEX_APP_SERVER_LOGIN_CLIENT_ID` — those decide where a single-use refresh POSTs and which client id is claimed.

Same class, also fixed: `CODEX_SQLITE_HOME` / `sqlite_home` moves the state DBs, so every profile's goals/logs/memories/state landed in one directory and the session home's durable links were never opened through — two accounts sharing one conversation history.

## 4. Decision 8's adopt makes an operator-side `codex login` revoke the captured chain

This is the one I'd most like your read on. After capture, `~/.codex/auth.json` is a symlink to `profiles/<A>/auth.json`. codex's `login` and `logout` are not in-place writers: both go through `clear_existing_auth_before_login` → `logout_with_revoke`, which **loads the stored auth through that symlink and POSTs its refresh token to the revoke endpoint** before deleting. So any operator-side `codex login` kills profile A's chain server-side, and no re-login of A can undo it.

Worse, my own refusal text was prescribing it — "already captured as codex profile 'A' … run `codex login` to mint a fresh chain". I've fixed the copy (name the hazard, say to remove the link first, and repeat the warning in the adopt success message), but the underlying shape is decision 8's, not mine: as long as the operator slot is a live link into a profile store, a normal codex command revokes that profile. If you'd rather the adopt were reversible by a first-class verb (`clauth codex detach <p>`, restoring the slot to a plain file) I'm happy to add it.

## 5. Decision 8's concurrency premise does not hold under `LinkMode::Fake`

"Concurrent codex sessions are allowed, and safe because they share one physical `auth.json`" is true under real symlinks. Under Fake the two flavors collapse to *different* bare stems (`codex-home`, `codex-home-isolated`), each holding its own copy converged from the same store — two carriers of one single-use chain, with no reload-and-skip between them since codex's own answer needs one inode.

Fixed by refusing the second *flavor* of one profile under Fake (same flavor is still free — that is one home). Tell me if you'd rather it converged harder instead of refusing.

## 6. The `sessions/` sync-back is write-only, and the thread DBs point into deleted homes

The runtime table has `sessions/` as a real per-home dir synced back to the store at teardown "so a rollout survives its session". Nothing ever restores it: `build_codex_home` only mkdirs a fresh empty one, and `sessions` isn't a durable entry. Meanwhile `state_5.sqlite` and `thread_history_1.sqlite` *are* durable and hold absolute rollout paths into per-session homes that no longer exist, and codex's listing scopes rollout paths under the sessions root — so prior threads are both invisible and prunable.

I did not change this, because making `sessions/` a symlink into the store (which would canonicalize to one stable path across sessions, the way the DBs already do) contradicts the table you wrote. It looks to me like the right fix; confirm and I'll make it part of phase 3.

Two smaller ones I fixed without needing a ruling: `archived_sessions/` was in no list at all, so archiving a thread deleted it at teardown; and codex's corrupt-DB recovery renames the *symlink* into `db-backups/`, leaving the corrupt file in the store forever and re-healing every session, so teardown now carries back a durable entry that has become a real file.

xingfanxia:
Implementation is up as #69 — the six-part series, 28 commits cut from `mommy` and rebased onto `eb6cbb8`, every commit building on its own.

Leaving this thread as the design one. The two items I asked about above are unchanged and still yours to rule on: the managed-config / MDM layer outranking `-c` (which is decision 6's premise), and the write-only `sessions/` sync-back whose fix would contradict the runtime table. Both are called out in #69's description so a reader of the diff hits them too.

uwuclxdy:
hi! dont think its worth going forward on this branch as #69 supersedes it and #51 will not be merged

uwuclxdy:
Closing: #69 merged with the series and the review's asks, so this design thread is done. The follow-ups the merge left (the Overview marker animation, the copy pass, codex name completion, `clauth list`'s codex rows, `clauth proxy`) are tracked maintainer-side.

PR #53, opened 2026-07-18
feat(claude): session-token split — static long-lived tokens end the refresh-race account deaths

### LINKED ISSUE #52: Serial "refresh token revoked" account deaths when long-lived Claude Code sessions coexist with clauth's refresher (+fix proposal: static session tokens)
## The failure

Over 2026-07-16..18 all three of my claude profiles serially died with `login for 'X' has expired: refresh token revoked or invalid` — each within hours of the previous, across different accounts. Re-login fixed each one for a day or two, then it died again.

## Root cause (verified on-device)

Anthropic OAuth refresh tokens are **single-use**: each refresh rotates the chain, and refreshing with an already-spent token trips server-side reuse detection, which revokes the entire chain.

The failure needs nothing exotic — just a user who (a) keeps Claude Code sessions open for days and (b) switches accounts with clauth:

1. A long-lived session holds its account's chain **in memory** and re-reads the credential store only at token boundaries (~8h). After a switch, it can keep rotating its *old* account's chain for up to a full token lifetime, writing rotations back into the one live slot.
2. clauth's standby refresher (and the pre-switch auth gate) refresh the *stored* copies of those same chains.
3. Two writers, one single-use chain: whoever refreshes second with the stale token kills the chain for everyone — the parked store copy AND the running session.

With N concurrent sessions across M accounts the collision is a matter of time; my daemon log shows the whole progression (live-slot stomps logged as `rotated 'X' but the live login diverged`, then the deaths).

## Fix proposal: split credentials — static session tokens (PR attached)

`claude setup-token` mints a **long-lived (~1yr) static token with no refresh token**. The PR lets a profile carry one in `<profile>/session-token.json`:

- switches install the **session token** into the live slot → sessions run on a token that never rotates, so there is nothing to race;
- the rotating OAuth pair in `credentials.json` becomes clauth-private, used **only for usage polling** (setup-tokens can't replace it — they carry only `user:inference` + `user:sessions:claude_code`, so `/api/oauth/usage` 403s; verified live), refreshed solely by clauth — single-writer restored.

Profiles without the sidecar behave byte-identically. Validated end-to-end against the real backend on a throwaway host: switch installs the token, real turns (including image reads) succeed on it, the usage pair survives byte-identical, nothing rotates.

Happy to adjust naming/UX — the sidecar is currently filled by hand; a `clauth login --setup-token` style capture flow could follow.

xingfanxia:
Closes #52 — root-cause analysis there.

> **Disclosure** (same as #45/#49/#51): implemented with an LLM (Claude) under my direction. Rebased onto current `mommy`; full suite green on this branch (**1111 tests, 0 failures**), clippy + fmt clean. Unlike the codex PR, this is claude-side and fully testable on your setup.

## What it does

A profile may carry a **static long-lived session token** (`<profile>/session-token.json`, minted by `claude setup-token`, no refresh token, ~1yr). When present:

- **Switches install the session token** (symlink + Keychain) via the new single concept `install_source_path` — sessions run on a token that never rotates, so N concurrent long-lived sessions can no longer race clauth's refresher on a single-use chain.
- **`credentials.json` becomes clauth-private, usage-only.** Setup-tokens carry only `user:inference` + `user:sessions:claude_code` — `/api/oauth/usage` 403s them (verified live) — so the rotating OAuth pair stays, but with exactly one writer: clauth.
- `classify_credentials_link` / `is_first_login` / snapshot / adopt / `clauth start` runtimes all resolve through the install source, so a live slot holding the token classifies `LinkedTo` and the divergence machinery stays dormant. Snapshot and the runtime watchdog never write the token over the usage pair; the rotation-persist path never Keychain-mirrors the usage chain over an installed token.
- `ensure_installable` returns `Ready` for session-token profiles — a stale/broken usage chain must not bench a switchable account — unless the token itself is clock-dead, which gates `Broken` with a re-mint hint.

**Profiles without the sidecar are byte-identical to before** (the install source falls back to `credentials.json`); additive only.

## Validation

- 4 new behavior tests (install-source preference, LinkedTo classification + snapshot no-clobber, switch installs the token and preserves the usage pair on disk, gate Ready under `auth_broken`).
- Real-backend end-to-end on a throwaway host, isolated `$HOME`: clauth switch links the live slot to the session token → real `claude -p` turn succeeds on it → an image-reading turn succeeds (setup-token scope is fine for images — they ride messages inline) → usage pair byte-identical after the turns → live token unrotated → the host's real `~/.claude` untouched (sha256). 7/7.
- Running in production on my machine since 2026-07-18; the account deaths stopped.

## Follow-ups (now included, second commit)

- **`clauth login <profile> --setup-token [--yes] [--model <id>]`** — guided capture of a `claude setup-token` mint into the sidecar (echo-off paste on a TTY, one stdin line for scripts; validated before anything is written, value never echoed/logged, 0600, stamped with the documented ~1yr lifetime). Additive and deliberately does NOT touch the live slot — takes effect on the next switch, so capturing can never sign a running session out.
- **Setup-tab `session` status row** — "static token · expires in ~Nd", WARNING inside 30 days, DANGER "token expired · re-mint: claude setup-token" (same wording the install gate logs).


uwuclxdy:
makes sense, but before this lands i'd want it to include:

- a `clauth login --setup-token` capture flow, so the sidecar isn't hand-edited
- backend detection of whether the installed token is long-lived, so clauth only skips the refresh-race handling when it actually is one. the TUI surfacing ("session token installed / expires in N days") can follow in a later pass

lmk if that works, or if you'd rather split the capture flow into a follow-up.

xingfanxia:
Pushed the two listed follow-ups into this PR (2022c4e, after a merge of current mommy): the `--setup-token` capture flow (echo-off paste / stdin pipe, validated, never touches the live slot — takes effect on the next switch) and the Setup-tab `session` row counting down to the re-mint. PR body updated; suite green on the branch, clippy clean.

xingfanxia:
Backend detection is in (2244a46): `session_token_status` classifies the sidecar by content — long-lived is defined by carrying NO refresh token (nothing to rotate = nothing to race, which is the split's whole point). A sidecar mis-filled with a rotating pair disengages the split entirely: `has_session_token` reads false, switches install `credentials.json` as if the sidecar weren't there, the install gate logs the mis-fill with the re-capture command, and the Setup-tab row shows "not long-lived (has a refresh token) · ignored" in DANGER. Also renamed the user-facing term to **long-lived token** throughout (CLI prints + the Setup row, now keyed `token`) — that's `claude setup-token`'s own wording, and "session token" reads like the opposite of a ~1yr static login; the on-disk `session-token.json` name is unchanged. So all three of your asks are now in this PR: capture flow, detection, and the TUI surfacing you were happy to defer. Suite green, clippy clean.

uwuclxdy:
Awesome, thank you! Will review and merge now

uwuclxdy:
merging now. pushed a few fix commits on top (claude-driven):

- `force_snapshot_active_credentials` was missing the `has_session_token` guard that `snapshot_active_credentials` got, so the divergence "overwrite" (and the CLI reconciled switch) could still clobber a long-lived profile's private usage pair. moved the guard down into the shared `snapshot_active_credentials_unchecked` sink so both paths are covered.
- the `token` row read `~0d` in WARNING for a token expired <24h ago instead of the DANGER expired hint (truncating `days`). gates on `now >= expiry` now.
- added `--setup-token` + login `--yes` to the bash/zsh/fish completions, plus `session-token.json` in the `SECURITY.md` file table.
- CI was red on two pre-existing things. the session-token tests didn't compile on windows (`HomeSandbox` import + `seed_relogin_scenario` were `#[cfg(unix)]` but used ungated). `rustfmt` 1.9.0 wanted a test line reflowed. both green now.

thanks for carrying this through all three follow-ups 🥺

PR #55, opened 2026-07-18
feat(scheduler): per-model weekly windows join the chain walk

### LINKED ISSUE #54: Auto-switch ignores per-model weekly windows — lands sessions on a "7d fable"-exhausted account
## The failure

Anthropic now caps Fable usage under its own per-model weekly window (`weekly_scoped` in the `/usage` `limits[]` array, label `7d fable`) — roughly **half** the aggregate weekly quota. The data layer already parses these windows and the TUI renders them, but the chain walk gates only on the aggregate `seven_day`.

That inverts into a stranding failure, observed live 2026-07-18 with three accounts:

| account | 5h | 7d | 7d fable |
|---|---|---|---|
| A | 23% | 18% | 12% |
| B | 0% | **65%** | **100%** |
| C | 17% | 67% | 90% |

To the aggregate-only walk, **B is the healthiest target** (5h idle, weekly 65%). But B refuses every Fable request until its weekly reset — a Fable session auto-switched onto B is dead in the water, while A sits idle.

The existing comment in `fallback.rs` documents the old stance: *"a per-model `weekly_scoped` limit at 100 blocks just that model, and the walk cannot know which model the next session will drive."* That premise was fine when per-model windows were rare; with Fable permanently available and capped at ~half the weekly, fable-heavy users hit this constantly.

## The fix (PR to follow)

No new config, no model detection — the walk stays model-agnostic but conservative:

- **Preference, not gate**: candidates clear on the aggregate gates AND every live `weekly_scoped` window rank above aggregate-clear-only members (scoped-clear outranks the freshness preference — a *known* model block beats a stale read's uncertainty). When no fully-clear member exists, acceptance falls back to today's behavior, so a fully model-blocked chain still rotates for the other models instead of halting.
- **Active trigger**: an otherwise-healthy active whose scoped window crossed the weekly line hops only when a fully-clear member exists — never ping-pongs between equally blocked members.
- **Scoped ≠ dead**: wrap-off `Off`, `soonest_resume`, and the recovery premise stay keyed on the aggregate cap (a fable-dead account still serves every other model).

Validated against the live shape above; 7 new walk tests pin it.

xingfanxia:
Closes #54.

Anthropic's "7d fable" per-model weekly window (~half the aggregate weekly quota) makes the aggregate-only walk land sessions on model-exhausted accounts — full story and live account shapes in #54.

## What this does (reworked per review — per-account toggles)

Two new toggles on each member's Fallback-tab card (both **on by default**), controlling what auto-switching checks for that account:

| row | on (default) | off |
|---|---|---|
| `weekly gate` | judged against the soft weekly line (stock behavior) | soft line ignored for this account; the **100% hard cap always blocks** (past the refusal cap it can't serve at all) |
| `scoped gate` | a per-model weekly window (e.g. "7d fable") past the line takes this account **out of rotation** | scoped windows ignored — keeps rotating here for other models |

| `weekly at` | follows the chain-wide `weekly limit` (Config tab stays the default) | — an override replaces the line for this account only; empty commit clears back; governs the aggregate AND per-model judgments; never the hard-cap ones; dimmed while `weekly gate` is off |

Hints on the rows state the current behavior and flip wording with the toggle. Persisted as `check_weekly` / `check_scoped` / `weekly_threshold` in the profile's config.toml — absent = stock behavior, so every existing config keeps its gating.

## Walk semantics

- Candidate accept (both walks): headroom AND clear per the member's own gates. Freshness stays a preference, not a gate — the any-freshness pass still always runs, so an exhausted active keeps its stale-but-viable escape. Chains whose accounts carry no scoped windows behave bit-identically.
- Active side: an otherwise-healthy active whose gated scoped window crossed the line hops only when a clear member exists (no ping-pong when every sibling is equally blocked). Mirrored in the UI one-shot and the scheduler walk.
- Scoped windows never count as full exhaustion: wrap-off `Off`, `soonest_resume`, and the recovery premise stay keyed on the aggregate cap — a fable-dead account still serves every other model. Recovery prefers gate-clear members and still relinks a model-blocked one when it's all there is (the chain is off at that point; anything serving beats nothing).
- New `ScopedSpent` blocked-reason chip: selector marker `◇` + detail pill naming the worst gated window ("7d fable 100% · other models ok"), so the card explains exactly why the walk skips the member. Gate off → no chip (chip and walk can't drift).

15 new/reworked tests (gate on/off per walk direction, hard-cap floor, active hop/stay/gate-off, recovery gates, chip precedence + gate-off, config round-trip, TUI toggle persistence + row rendering); full suite green, clippy clean.

Claude-side only — reproducible against the real `/usage` shape on any account with a `weekly_scoped` entry.


uwuclxdy:
i see where you're coming from but i'd rather have this a toggle per account on the fallback tab. the reason is that users may still want to use that account with other models or exclude only the maxxed out fable account from rotation. i hope you know what i mean.

can you implement these two toggles (feel free to suggest better names for them) in the fallback tab, per account?
- "weekly usage   ○─" with dynamic hint: "(don't) check weekly usage when auto-switching")
- "scoped usage   ○─" with dynamic hint: "(don't) check scoped usage when auto switching; e.g. 7d fable"

also, what do you think about moving the `weekly limit` setting from config tab into fallback tab, per account?

xingfanxia:
Done in 3d312a1 — both toggles are on each member's Fallback-tab card now, PR body rewritten to match.

Naming: I went with **`weekly gate`** / **`scoped gate`** instead of "weekly usage" / "scoped usage" for two reasons — (a) the card already has a `5h usage` row that *displays* usage, so "weekly usage ─●" read like a data row rather than a switch; (b) both fit the existing 11-col key column ("last resort" width), so the value column stays aligned with the Config tab. The row hints carry your wording and flip with the state, e.g. scoped off → "per-model weeks aren't checked; stays in rotation for other models". Trivial to rename if you prefer the originals.

Two semantic calls you may want to veto:

1. **Defaults: both ON.** Weekly-on is stock behavior. Scoped-on means the #54 fix works out of the box — an account whose "7d fable" is spent drops out of rotation until its week resets or you flip its gate off. That's a behavior change for chains that carry scoped-capped accounts today (they'll rotate around them instead of onto them); the gate-off toggle is exactly the escape for "I run other models on it".
2. **`weekly gate` off still respects the 100% hard cap.** Past the refusal cap the account can't serve anything, so switching to it only manufactures errors — the toggle drops the soft line, never the cap.

Also added a `ScopedSpent` chip (◇ marker + "7d fable 100% · other models ok" pill) so the card explains why the walk skips a gated member; gate off → no chip.

On moving **`weekly limit`** from the Config tab to per-account: I'd make it a per-account *override* rather than a move — keep the Config-tab value as the chain-wide default and let a member's card override it, exactly the `rotate at` / default-threshold pattern. A full move loses the one-knob chain-wide tune (changing the line = N edits) and needs a config migration; an override is additive and existing configs keep working. Happy to do that as a follow-up PR (or fold it in here if you'd rather take it in one piece).

xingfanxia:
Folded the weekly-limit half in too (b12ec28), per the override-not-move shape from my last comment: each member card gains a `weekly at` row — Config-tab `weekly limit` stays the chain-wide default, a member override replaces it for that account only (empty commit clears back), and the one per-account line governs both the aggregate 7d judgment and its per-model windows. Hard-cap judgments (sinks, halts, dead-for-days chip, recovery premise) deliberately ignore it — no per-member taste may soften "can this account literally serve". Row is dimmed/inert while the member's weekly gate is off. PR body table below still stands; suite green, clippy clean. The two default questions from my previous comment (gates both ON; weekly-gate-off keeps the 100% cap) are still yours to veto.

uwuclxdy (changes requested):
Thanks for reworking this to per-account gates, the card surface reads clean.

One correctness issue before it can land, plus a CI thing.

**Root cause:** making `weekly_blocked` / `is_exhausted` resolve the per-member line through `member_weekly_line` (`weekly_threshold.unwrap_or(chain_soft)`) silently reinterprets every pre-existing caller that passed `WEEKLY_HARD_BLOCK_PCT` as a literal hard-cap check. `member_weekly_line(p, 100)` with an override of `50` returns `50`, so those callers now do a soft judgment. You added `is_exhausted_hard` / `weekly_hard_blocked` for exactly this and wired `soonest_resume` + `blocked_reason`, but four hard-cap sites still go through the soft path:

- `fallback.rs:942` and `:947` (the `next_target` serving-sink passes). A `last_resort` sink with a `weekly at` override below 100 and aggregate 7d sitting between the override and 100% (5h idle) still serves for free, but now reads as "can't serve", so the sink pass skips it and the spend pass pays a real-money account instead. The scheduler twin `next_auto_switch_target` (`:1150`) passes the literal 100 via `is_exhausted_from_store`, so the daemon parks free; the TUI-side `auto_switch_if_needed` spends. The two twins diverge on real money.
- `tui/render/usage.rs:144` (`weekly_hard` diagnostic) flags an account "dead for days" when it is below the true 100% cap and still serving. `usage.rs` isn't even touched by the PR, the helper change reaches back into it.
- `tui/app.rs:7145` (`update_banner`'s `any_spent`) shows a false "all accounts spent" DANGER banner. The comment right above it says the evidence has to be the hard cap because a soft-blocked member still serves.

All four want `is_exhausted_hard(p)` / `weekly_hard_blocked(p)` in place of the `..(p, WEEKLY_HARD_BLOCK_PCT)` form.

`weekly_override_never_softens_the_hard_sink_and_halt_judgments` only drives `next_auto_switch_target` (the twin that's already correct), so the buggy `next_target` path stays green. Worth a case that runs `next_target` / `auto_switch_if_needed` with an override on a `last_resort` sink and a spend-armed sibling, which is the exact divergence.

**CI:** `cargo fmt --check` reds the tree. A bunch of the new `ChainMember` test literals have two fields on one line (`scoped_line: 98.0,            check_scoped: true,`) and rustfmt splits them, so the fmt leg would fail.

Smaller stuff (the `weekly at` edit action missing from the `?` help modal, the `chain.rs` module doc dropping the `weekly at` row, a couple of doc-sync bits) I'll fold into a followup rather than block on here.

lmk what you think

xingfanxia:
All four sites fixed in eb8394a — the `next_target` serving-sink passes, the usage-tab `weekly_hard` diagnostic, and `update_banner`'s all-spent evidence now go through `is_exhausted_hard` / `weekly_hard_blocked`, so no member line ever reaches a hard-cap judgment. Added the exact-divergence tests you described: `next_target` with a weekly override on a `last_resort` sink plus a spend-armed sibling (both facets — sink-as-target parks free instead of paying, sink-as-active stays put). Verified they fail on the pre-fix code at precisely the paying step before going green.

fmt: `cargo fmt` reflowed the two-fields-per-line `ChainMember` literals (plus a few it flagged nearby); `cargo fmt --check` is clean on the branch now. Also folded in two of the small bits so your followup shrinks: the `?` help modal gained the `weekly at` entries and the `chain.rs` module doc names the row again. Suite 1161 green, clippy clean.

uwuclxdy (changes requested):
Thanks for the rework. The scoped-gate core is good work and it does close #54, the walk tests around it are well built. Needs one more round before it can land though.

The one I'd fix first: `member_scoped_line` deliberately ignores `check_weekly`, so with `weekly gate` off and `scoped gate` on a stored `weekly_threshold` keeps judging every per-model window. The row meanwhile renders dimmed, `run_fallback_row` refuses to open the editor, and the hint says the line isn't checked for this account. There's no way back to that value from inside the TUI. imo the fix is one line, `member_scoped_line` honoring `check_weekly` the same way `member_weekly_line` does, so one gate means one thing. The render layer claiming a state the walk doesn't act on is the specific thing the chip contract at `fallback.rs:540` exists to prevent.

Second, this wants a rebase rather than a merge. `mommy` moved twice under the branch: `ROWS_BEFORE`/`PILL_LINES` are gone, replaced by a computed `rows_start` plus a `rows_start + row_idx < inner.height` bounds guard, so the branch's `chain.rs:237` cursor math is unguarded on a base that already fixed exactly that. Separately, `0fde6b6` consolidated `⊖` to cover two arms splitting on hue alone, which makes `◇` a tenth glyph going the other direction. Does `ScopedSpent` need its own shape, or could it ride `⊘` at a different hue?

<details>
<summary><b>Full findings</b></summary>

### Blocking

1. **`weekly gate` off leaves the override live.** `fallback.rs:223` (`member_scoped_line`), `tui/app.rs:4257`, `tui/render/chain.rs:427`. Covered above.
2. **Rebase onto `mommy`.** Conflicts on `fallback.rs`, `tui/render/chain.rs`, and four test files. Covered above.
3. **Two of the four hard-cap sites `eb8394a` fixed have no regression test.** `update_banner`'s `any_spent` (`tui/app.rs:7138`) and the usage-tab `weekly_hard` diagnostic (`tui/render/usage.rs:141`). `all_spent_banner_fires_at_the_weekly_hard_cap` looks like it covers the first, but its `blank_profile("a")` fixture carries no override, so the pre-fix `member_weekly_line(p, 100.0)` resolves to `None.unwrap_or(100.0)` = 100 and the test is green against the buggy revision too.
4. **No fixture combines `check_weekly = false` with a non-empty `weekly_scoped`.** All six gate-off sites checked: `tests/inline/fallback.rs:2892` uses `both_windows` with no scoped entries, `:3121` passes `usage_with_scoped(0.0, 99.0, vec![])`. Replacing `member_scoped_line`'s body with `member_weekly_line(profile, chain_soft)` keeps the suite green. This is the same gap as finding 1: nothing pins the invariant, so nothing caught it.
5. **`fully_clear_target` has zero coverage**, along with the `auto_switch_if_needed` scoped branch that is its only caller. The scheduler twin got four tests. `tests/inline/actions.rs:186` already drives `auto_switch_if_needed` under `HomeSandbox` if that helps.
6. **The gate default is never proven at the load boundary.** Mutating `profile.rs:1135` `config.check_weekly.unwrap_or(true)` to `unwrap_or(false)` leaves the suite green. `profile_config_usage_gates_default_unset` stops at `ProfileConfig`, and the only `load_profile` test asserts both gates `false`, which the mutation satisfies. `weekly_switch_threshold_out_of_band_resets_to_default_at_load` is the right shape, it writes real TOML through the real boundary.

### Wanted before merge

7. **`weekly at` accepts a band its own default rejects.** `parse_weekly_override` (`tui/app.rs:4355`) routes through `parse_threshold` for `0..=100`; the chain-wide `weekly limit` it overrides uses `parse_weekly_pct` (`:4100`) for `50..=100` plus `is_finite`. Typing `0` permanently weekly-blocks that member with no warning. `parse_weekly_pct` is the call to make here.
8. **`weekly_threshold` should reset, not clamp, at load.** `profile.rs:1125` clamps. The chain-wide line it overrides deliberately resets an out-of-band hand-edit to the default instead, so a `40.0` reads as `98` and not `50`. That's a written invariant on my side, the rationale being that a fraction-vs-percent typo like `0.98` clamps to `0.98` and silently blocks the account from about 1% into its week, chipped as a plausible-looking `WeeklySoft`.
9. **The scoped-active trigger sits above the `last_resort` parking guard.** `fallback.rs:1109`, UI twin `:1325`, guard at `:1136`. It's the only path that relinks a healthy active, and it hops off a pinned sink. Chain `[A(sink, 7d 60%, fable 100%), B(clear)]` oscillates on roughly a 5h period: hop to B, B's 5h exhausts, the serving-sink pass returns to A, B's 5h resets, hop again. A pinned sink should stay parked.
10. **`blocked_reason` inlines the scoped predicate** at `fallback.rs:604` instead of routing through `scoped_weekly_blocked_info`. The function's own contract two hundred lines up says it's never a second opinion, and an inlined copy is how the chip and the walk drift apart.
11. **The member card clips at 40x24.** Three new rows on a card that doesn't scroll: 5 header lines + 7 rows + a 2-line tooltip fills the detail pane exactly, and any blocked member's 2-line pill pushes `max spend` and `remove` off with no scrollbar to reach them. It fit at 4 rows.
12. **One control, three names.** Row key `weekly at` (`chain.rs:679`), action menu `edit weekly line` (`app.rs:737`), help modal `edit weekly at` (`modals.rs:529`).

### Notes, no need to action

- `weekly_threshold = nan` is valid TOML, survives `clamp`, makes `Some(NaN) != Some(NaN)` fire the config rewrite, renders as `NaN`, which is *not* valid TOML, and then `load_profile` fails on the file it just wrote. `fallback_threshold` and `bell_threshold` do the same thing, so it's my pre-existing bug rather than yours. I'll fix all three to follow `max_auto_spend`'s `is_finite` guard. Happy for you to fold `weekly_threshold` in here if you'd rather.
- `fully_clear_target` (`fallback.rs:1218`) re-derives `active`/`chain`/`active_idx`/`skip` and the accept predicate verbatim from `next_target:887`. Both copies are correct today, it's just two things to keep in lockstep by hand.
- `chain.rs:690` hand-rolls the `   default: ` span that `global_config.rs:400` already provides as `default_reminder`.
- A few test cleanups worth a pass if you're in there anyway: `weekly_threshold_round_trips_and_clamps` never clamps and never calls `load_profile` where the clamp lives; the `ScopedSpent` precedence chain only has a fixture for its edge against `WeeklySoft`, so moving it above `FiveHour` stays green; `last_resort_hint_wraps_on_a_narrow_pane` still passes `cursor = 1` commented as `FALLBACK_ROWS[1]`, which is now `WeeklyAt`.
- CI hasn't run on the branch, so I can't confirm the green suite from here. Nothing on you, just noting why I checked the tests by hand.

</details>

Internal design docs go stale on merge, those are gitignored so there's nothing for you to update. I'll reconcile them on my side.

thoughts?

xingfanxia:
Round pushed: rebased onto `mommy` (the branch now replays clean on `d68fb2e`'s line) + three fix commits on top — `17db09f` / `8bf9bd2` / `51338fb`.

**The first one:** `member_scoped_line` now rides `check_weekly` — with the gate off the stored override is inert and scoped windows judge at the chain line (scoped judgment itself stays on; `check_scoped` remains its gate). I pinned both directions, since there were two ways to get this wrong: the suite goes red if the override leaks past the gate, and also red if gate-off routes the scoped line to the hard cap the way `member_weekly_line` does — replacing its body with `member_weekly_line(profile, chain_soft)` no longer survives (your finding 4's fixture gap: there's now a `check_weekly = false` × non-empty `weekly_scoped` fixture on the walk AND the chip).

**The rebase:** done as a rebase, not a merge. Your `rows_start` rework made the cursor-math port trivial — the scroll fix below reads straight off it.

By finding:

1. Fixed as above.
2. Rebased; `chain.rs` cursor math now sits on the `rows_start` + bounds-guard base.
3. Both hard-cap sites have discriminating regressions now: the banner test carries a `weekly at 90` override at 7d 95 (your exact point — the old `blank_profile` fixture's member line WAS the cap, so the folding revision passed it), and the usage-tab teach has the same override-shaped pin through a full-frame render.
4. Covered by the fixtures in 1.
5. `fully_clear_target` + the `auto_switch_if_needed` scoped branch have direct coverage: skip scoped-blocked/broken, `None` when everything's blocked, and the real one-shot landing the switch under `HomeSandbox` (thanks for the `actions.rs:186` pointer).
6. The gate default is proven at the load boundary now — empty config.toml through `load_profile` loads both gates ON; mutating `unwrap_or(true)` to `unwrap_or(false)` goes red.
7. `parse_weekly_override` routes `parse_weekly_pct`; tooltip says 50-100.
8. Reset-not-clamp at load, mirroring the chain-wide line: `0.98`, `40`, `150`, and `nan` all reset to unset, pinned through real TOML + `load_profile`. That also folds your `weekly_threshold` NaN note in — the filter kills non-finite before it can round-trip.
9. Both walks keep a pinned sink parked (`!active.last_resort` on the trigger), pinned scheduler-side and through the UI one-shot. Your oscillation trace was right — the serving-sink pass would have kept dragging it back on a ~5h period.
10. The chip's judgment routes through a new `worst_scoped_window`, which `scoped_weekly_blocked_info` now IS (`is_some()` over it) — one predicate, the display just names the window it found.
11. The focused card scrolls the cursored row (plus its tooltip) into view, capped at the content bottom; unfocused keeps the top anchored. The typed-field caret subtracts the same scroll. Added a 40x24 test walking the cursor to `remove` with a blocked member's pill block in the header. (The upstream clipped-caret test flipped to pin the new contract: the typed row scrolls into view and the caret lands on its rendered position.)
12. `edit weekly line` → `edit weekly at` everywhere.

On the glyph question: agreed, and done — `ScopedSpent` rides `⊘` at warning hue. Hue splits scope exactly like `⊖` splits disabled/canceled: danger = the aggregate week (dead for days), warning = one model's week (still serves the rest). `◇` is gone.

Test-cleanup notes: `weekly_threshold_round_trips_and_clamps` renamed to drop the false "clamps" claim (the reset test owns the boundary now), the `ScopedSpent` precedence chain gained its `FiveHour` edge (moving it above goes red), and `last_resort_hint_wraps_on_a_narrow_pane` derives its cursor from `FALLBACK_ROWS` instead of the stale hardcoded 1. The `fully_clear_target`/`next_target` verbatim duplication I left as-is per your note — it now at least has its own tests to keep the copies honest.

Every new discriminator was mutation-checked red against the revision it guards before committing. Local gates: 1368 tests green, clippy clean, fmt clean.

PR #58, opened 2026-07-21
fix(switch): a clauth-owned live symlink is never "unsaved credentials"

xingfanxia:
## The bug

Capturing a setup-token sidecar for the **active** profile (`clauth login <p> --setup-token`, #52/#53) flips `install_source_path` from `credentials.json` to `session-token.json` — while the live slot still holds the old symlink pointing at `credentials.json`. From that moment:

- `classify_credentials_link` reads **Diverged** (correct: the link no longer points at what a switch installs),
- but `active_diverged_unsaved` treats every Diverged non-shell state as an unsaved login needing protection, so
- every unattended switch fails `active '<p>' has unsaved credentials; resolve in the TUI` until its retry TTL, switch-off is skipped, and the TUI raises the divergence prompt about credentials that are fully saved.

Hit live 2026-07-21 on the macOS fork while repairing a mis-filled sidecar: a queued switch retried the failure loop until `gave up switching`.

## The fix

A clauth-owned symlink's content is a profile store **by construction** — re-pointing the link loses nothing, there is nothing unsaved to protect. New `claude::live_login_is_clauth_symlink()` exempts that state in both `active_diverged_unsaved` gates (daemon + TUI parity). A regular-file live login (a real CC `/login`) is gated exactly as before.

Regression test walks the full transition: LinkedTo before the capture → Diverged-but-not-unsaved after → regular-file and absent live slots still classify as before.

1237 tests green, clippy clean, fmt clean.

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -9,9 +9,8 @@ use std::sync::Arc;
 use anyhow::{Context, Result, bail};
 
 use crate::claude::{
-    ClaudeEndpoint, LinkState, apply_profile_to_claude_settings, classify_credentials_link,
-    clear_claude_credentials, force_link_profile_credentials, force_snapshot_active_credentials,
-    is_first_login, link_profile_credentials, live_credentials_are_shell,
+    ClaudeEndpoint, apply_profile_to_claude_settings, clear_claude_credentials,
+    force_link_profile_credentials, force_snapshot_active_credentials, link_profile_credentials,
     live_diverged_and_unsaved, managed_env_key_label, read_claude_credentials,
     read_claude_endpoint_config, snapshot_active_credentials,
 };
@@ -77,20 +76,21 @@ pub(crate) fn switch_profile(config: &mut AppConfig, name: &str) -> Result<()> {
         // first-login), so dropping it would strand a fresh `/login` chain — keep
         // the non-force refuse-guard there. Every other state is captured or
         // adoptable by the snapshot below, so force the relink: on macOS the live
-        // `.credentials.json` is always a regular-file Keychain mirror of the
-        // active account and thus legitimately differs from the target, which the
-        // non-force guard's live-vs-target byte check would wrongly reject on every
-        // switch. (Interactive callers already route a real divergence to the
-        // reconcile path, so this branch is only reachable uncaptured via the
-        // scheduler — where refusing, not dropping, is the safe outcome.)
-        // A logged-out shell holds no login to strand, so it forfeits the
-        // refuse-guard (which would otherwise wedge the switch on an empty file).
+        // `.credentials.json` is a regular-file Keychain mirror of the active
+        // account, so it legitimately differs from the target, which the non-force
+        // guard's live-vs-target byte check would wrongly reject. The SAME
+        // predicate the defer/banner gates use — `live_diverged_and_unsaved` —
+        // decides here, so a login already saved in the store (the mirror, a
+        // clauth symlink) forces the relink even once a sidecar capture flips the
+        // install source and makes classify read Diverged over it; without that
+        // exemption the guarded link byte-rejects the macOS mirror and the switch
+        // fails "unsaved credentials" though nothing is unsaved. (Interactive
+        // callers already route a real divergence to the reconcile path, so this
+        // branch is only reachable uncaptured via the scheduler — where refusing,
+        // not dropping, is the safe outcome.) A logged-out shell holds no login to
+        // strand, so it too forfeits the refuse-guard.
         let uncaptured_relogin = match config.state.active_profile.as_deref() {
-            Some(active) => {
-                matches!(classify_credentials_link(active)?, LinkState::Diverged)
-                    && !is_first_login(active)?
-                    && !live_credentials_are_shell()
-            }
+            Some(active) => live_diverged_and_unsaved(active)?,
             None => false,
         };
         snapshot_active_credentials(config)?;
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -11,8 +11,9 @@ use anyhow::{Context, Result, bail};
 use crate::claude::{
     ClaudeEndpoint, LinkState, apply_profile_to_claude_settings, classify_credentials_link,
     clear_claude_credentials, force_link_profile_credentials, force_snapshot_active_credentials,
-    is_first_login, link_profile_credentials, live_credentials_are_shell, managed_env_key_label,
-    read_claude_credentials, read_claude_endpoint_config, snapshot_active_credentials,
+    is_first_login, link_profile_credentials, live_credentials_are_shell,
+    live_diverged_and_unsaved, managed_env_key_label, read_claude_credentials,
+    read_claude_endpoint_config, snapshot_active_credentials,
 };
 use crate::lock::with_state_lock;
 use crate::lockorder::RankedMutex;
@@ -140,11 +141,7 @@ pub(crate) fn switch_profile_cli(config: AppConfig, canonical: &str) -> Result<(
     // exempt: capturing its blank tokens would destroy the outgoing profile's
     // stored login.
     let reconciled = match outgoing.as_deref() {
-        Some(active) => {
-            matches!(classify_credentials_link(active)?, LinkState::Diverged)
-                && !is_first_login(active)?
-                && !live_credentials_are_shell()
-        }
+        Some(active) => live_diverged_and_unsaved(active)?,
         None => false,
     };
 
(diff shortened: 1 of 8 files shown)


uwuclxdy (changes requested):
the mechanism is right and the safety reasoning holds. a clauth-owned symlink's target is always a profile store, and `snapshot_active_credentials` still skips the `Diverged`-not-first file, so re-pointing can't lose anything unsaved. no notes there.

two things before merge though.

**the exemption only reaches 2 of the gates.** `!live_login_is_clauth_symlink()` went into the daemon + TUI `active_diverged_unsaved`. three more spots run the same `Diverged && !is_first_login && !live_credentials_are_shell` triple, each still misfiring on the stale symlink:
- `switch_profile_cli` (`actions.rs:160`): spurious `'X' has a newer login in ~/.claude ... [Y/n]` prompt
- `switch_profile_noninteractive` (`actions.rs:295`, the MCP `switch` tool): bails `has a login clauth hasn't saved` when no divergence default is set
- `poll_credentials_divergence` (`tui/app.rs:7343`): repaints the divergence banner every second, which is the TUI symptom from your description. the switch modal is fixed but the 1Hz nag isn't

`switch_profile`'s `uncaptured_relogin` (`actions.rs:105`) is genuinely fine, a symlink repoints either way there. imo the clean fix is one `claude::live_diverged_and_unsaved(active)` that all the callers hit with the term added once. the two you edited collapse into it too, so it can't drift out of sync again.

**no regression test for the actual defer.** the new test pins `live_login_is_clauth_symlink`, not the switch-that-should-now-proceed, so dropping the `&&` term stays green. `drain_pending_switch_proceeds_over_a_logged_out_shell` in `tests/inline/daemon_mod.rs` is the exact shape. a `..._over_a_stale_clauth_symlink` sibling would red on the real bug.

non-blocking: this is unix-only by construction. on Windows the live slot is a copy so `is_symlink` is always false and the install-source flip still trips the gate. fine to leave if setup-token-on-active isn't reachable there, just flagging.

thoughts?

xingfanxia:
Round pushed as an append (`b53226d`, no force) so the delta since your review stays a clean diff.

**The gates:** one `claude::live_diverged_and_unsaved(active) -> Result<bool>` now owns the composite triple + the symlink term, and the daemon gate, the TUI gate, `switch_profile_cli`, and `switch_profile_noninteractive` all route through it — the two round-1 inline copies collapsed into it, so it can't drift again. Signature note: it takes just the name, because four of the five sites hold nothing precomputed; `classify` and `is_first_login` share the same fallible install-source prefix internally, so the composite costs no extra disk pass.

One deliberate deviation from "all five through the composite": `poll_credentials_divergence` keeps its early-return cascade and applies `live_login_is_clauth_symlink()` as its own early return instead. That site interleaves the first-login *adopt* side effect between the triple's terms, so routing it through the composite would either re-order the adopt or triple the poll's 1Hz disk reads. The symlink judgment itself is still single-sourced — the poll adds the one atomic check, not a copy of the triple. If you'd rather eat the extra read for uniformity, happy to flip it.

**The regression test:** `drain_pending_switch_proceeds_over_a_stale_clauth_symlink` — the `drain_pending` sibling you asked for — plus `divergence_poll_ignores_a_stale_clauth_symlink` covering the 1Hz-banner site. Both mutation-checked: deleting the symlink term from the helper reds the first, neutering the poll's guard reds the second.

On the Windows note: agreed it's unix-only by construction, and setup-token-on-active shouldn't be reachable there (the live slot is a copy, so the stale-symlink state this PR fixes can't arise — the install-source flip tripping the gate on Windows is the pre-existing divergence semantics doing its job on a real divergence). Left as-is.

Gates: 1239 tests green, clippy clean, fmt clean.

uwuclxdy (changes requested):
Round 2 lands both round-1 asks. The shared `live_diverged_and_unsaved`, the `drain_pending` sibling that reds when the term drops, the poll keeping its adopt side effect single-sourced. I traced that leaving `switch_profile`'s `uncaptured_relogin` raw is correct too, a symlink repoints identically under the non-force and force link and both do the Keychain write. No notes there.

One thing neither of us caught last round, and it's on macOS, where this was found.

The exemption keys on `is_symlink()`, but macOS mostly isn't a symlink. CC rewrites `~/.claude/.credentials.json` as a regular-file mirror of the Keychain after every run (the clobber `classify_link_at` already documents on its not-our-symlink branch). So the moment `claude` runs, the live slot is a regular file holding the pre-flip `credentials.json` OAuth, the install-source flip still classifies `Diverged`, `is_symlink()` reads false, and the switch defers / banner nags exactly as before. The 2026-07-21 repro was in the window before CC re-clobbered, so the fix holds there, but the steady state on macOS after any session is uncovered. Same misfire, macOS again.

The claim that clears the symlink is about content: the target is a saved profile store, so re-pointing loses nothing. Nothing about it needs the link to be a symlink. It's your Windows note from round 1, except reachable on macOS instead of hypothetical. So imo the last term wants to be content-based: exempt when the live login already matches `active`'s stored `credentials.json` (the pre-flip source), which covers our symlink, the macOS mirror, and the Windows copy in one predicate and drops the unix-only footnote. The three tests stay, we'd add a regular-file case so the macOS shape actually reds.

I can't test on macOS so I'm reasoning this from the classify comment plus the path, not a device repro, but I'm fairly sure it recurs there. Thoughts?

xingfanxia:
Good catch — you reasoned it right without the device. Fixed as an append (`e66b9cb`, no force).

`live_login_is_clauth_symlink` → `live_login_is_stored(active)`: a login is saved either **by content** (the live OAuth access token matches one of the profile's stored credential files) or **structurally** (a clauth-owned symlink, whose target is a store by construction). The content half is what you asked for — cross-platform, so it covers the macOS regular-file mirror and the Windows copy in one predicate and drops the unix-only footnote. I kept a symlink structural clause too, but only for the one case content can't reach: a **dangling** clauth symlink (its store file removed under the live link) — there's no login there to protect, so it should still exempt, and reading through it to compare content would fail. In the common readable case the content half carries the symlink anyway (reading through it yields the store file), so the mutation test that reds the macOS shape reds it whether the live slot is a symlink or a regular file.

One deviation from your exact suggestion: I compare against **both** stored files (`credentials.json` AND `session-token.json`), not just `credentials.json`. The flip is bidirectional — a sidecar capture moves the install source `credentials.json` → `session-token.json`, but a sidecar *removal* moves it back, leaving the live slot holding the static mint. Checking both keeps both directions saved; checking only `credentials.json` would re-defer the removal direction. Same cost, strictly wider.

**The part your macOS reasoning also surfaced, one layer down.** The defer gate was only half of it. `switch_profile`'s `uncaptured_relogin` ran the same `Diverged && !first-login && !shell` triple *without* the exemption (round 1 we both signed off on it as "fine" — correct for a symlink, where the guarded link skips its byte-check; **not** for a regular file). So on the macOS mirror, once the flip made classify read Diverged, the gate would now pass but `switch_profile` took the guarded-link path and byte-**rejected** the switch — "unsaved credentials" from the other side, the switch never lands. It routes through `live_diverged_and_unsaved` now too, so a login already in the store forces the relink. Both sites, one predicate.

I found that second site only because I wrote the fix as an end-to-end daemon switch test rather than a unit test on the predicate — the gate went green but the switch still failed, which is what a device repro would have shown you. So the regression coverage is: a cross-platform regular-file-mirror case on the helper, on the daemon drain-switch, and on the 1Hz poll — all three red without the content half, and the daemon case also reds if `uncaptured_relogin` keeps the raw triple. The three symlink tests stay (now through `live_login_is_stored`), plus a dangling-symlink assertion for the structural clause.

Gates: 1242 green, clippy clean, fmt clean.

uwuclxdy (approved):
This is it. The content-based rework lands the macOS ask cleanly. The `uncaptured_relogin` reroute closes the second site I waved through as "fine to leave raw" in R1/R2, which it wasn't: `link_profile_credentials` only skips its byte-check on a symlink, so the regular-file mirror hit the raw triple → non-force → byte-reject ("unsaved credentials" from the other side). Routing it through `live_diverged_and_unsaved` forces the relink on a stored login. Good that the end-to-end daemon test is what surfaced it rather than a unit test on the predicate.

Checking both stores instead of just `credentials.json` is the right call imo. The flip is bidirectional and a `NotLongLived` sidecar can still sit on disk, so one-store would re-defer the other direction. Strictly wider, same cost.

Same caveat as your side: I can't run the macOS path either, so the Keychain-mirror behavior is reasoned from the `classify_link_at` comment rather than a device repro. The non-`cfg(unix)` regular-file twins are the right way to pin the shape on CI regardless.

lgtm ✅

Issue #60, opened 2026-07-30
discussion: opt-in auto-return / greedy re-pick — recovery while a profile is still active


alidarvishi14:
## The gap

The fallback chain is one-directional. It switches *off* an exhausted account, but recovery (`scan_recovery` / `find_recovered_member`) only runs once `active_profile` is `None` — i.e. after switch-off-all. So the common two-account pattern has no automatic **return**:

> primary account → spill to the backup when the primary's 5h maxes → **come back** when the primary's 5h resets.

Today you ride the backup until it also trips, even after the primary is healthy again.

## Proposal — opt-in `auto_return` / greedy mode (default off)

Let the daemon switch *toward* a preferred account while one is still active, with anti-thrash guardrails (a switch moves every live session, so disruption is the whole design constraint):

- **preferred = most weekly (7d) headroom**, made **sticky** — only re-designated when the other account leads by a margin. Load-balances the scarce, days-to-recover weekly bucket without flapping between near-equal accounts.
- **offload** to the sibling only when the preferred's 5h crosses a high mark (~92%); **return** only once it falls back under a lower one (~70%). A hysteresis *band*, not a single threshold, so it can't oscillate at the boundary.
- **cooldown** between switches; skip while `pending_switch` is set or the feed is stale; never triggers switch-off-all (no sign-out).

Net behavior: 5h is treated as a short-term throttle you route around; the weekly bucket is balanced across accounts over the week; you spend most of your time on the account you actually prefer.

## Status

I validated the policy end-to-end against a live two-account Max setup with a small `status.json`-driven sidecar — it reads the daemon feed and acts through `clauth <profile>`, so **zero extra API calls** and it composes with the daemon (clauth's own file-lock serializes the switch). Hysteresis + cooldown behave as intended across offload → hold → return.

Happy to port it into the scheduler as an opt-in flag matching the existing lock/race discipline, if it fits the roadmap. Flagging as discussion first per #27 — and is any of this already on your path with the 7d/​scoped work in #38/#55?

uwuclxdy:
This seems like a good feature! Would it be possible to make it use the fallback chain order somehow, maybe as reverse walk through the chain? LMK if that's a good idea or if it should be just one preferred account that clauth returns to with a toggle in the TUI per account.

alidarvishi14:
Thanks! I went a different direction on the metric — happy to reconsider. Put it up as #61.

[...]

PR #61, opened 2026-07-30
feat(fallback): opt-in greedy re-pick — switch toward the highest weekly burn rate

alidarvishi14:
> **operator asked:** implement the greedy re-pick from #60 in Rust and open a PR from the fork once it's stable. My operator specifically wants the *rate-based* metric below (their idea), not a chain-order walk.

Adds an opt-in **greedy re-pick** so the fallback daemon can also switch *toward* a healthy account instead of only ever switching *off* a spent one — scored by each account's **sustainable weekly burn rate**.

The chain walk today is one-directional: `next_auto_switch_target` leaves an exhausted active but never returns, and `scan_recovery` only runs after switch-off-all. So "primary account, spill when its 5h maxes, come back when it resets" has no automatic return. This is #60.

With `greedy_switching` on (default off), the healthy-active branch of `next_auto_switch_target_with_usage` — previously an unconditional `None` — picks the clear, non-sink member with the highest **burn rate** and switches to it once the active's own rate falls below `GREEDY_RATE_HYSTERESIS` (0.8) of it:

```
rate(account) = (weekly line − 7d utilization) / hours until the 7d window resets
```

i.e. how much of its weekly budget it can spend per hour and have it last exactly to the reset. Design, argued in the code:

- **Rate, not raw headroom or chain order.** It folds in *time-to-reset*: an account at 40%/2-days-left sustains more per day than one at 55%/3-days, and one near its reset is "richer per hour" because its budget refreshes regardless. Leaning on the highest-rate account draws its rate down toward the field, so the daemon **equalizes burn across accounts (water-filling)** and keeps each alive to its own reset. The "return after a 5h offload" falls out for free — the offloaded-from account re-enters `clear` once its 5h resets and is usually the higher-rate one again.
- **5h is a pure gate, never scored.** A 5h-capped member is already excluded by the existing `clear` predicate, so greedy never chases a short-term throttle that refills on its own.
- **Multiplicative hysteresis.** The active is left only when `active_rate < 0.8 × best_rate` — a band (rates within 0.8–1.25× stay put), so two near-equal accounts don't trade the login as their windows drift. No persistent state; an equal-rate rival never displaces the active.
- **Reuses the walk's `skip`/`clear`/`fresh` closures**, so candidate eligibility is identical to the rest of the function; the single-store-lock-per-evaluation guarantee (`..._takes_store_lock_exactly_once`) is untouched. `last_resort` members are never targets.

Also adds a `greedy re-pick` toggle to the Config tab's auto-switch band (mirroring `burn-aware`) and a README bullet.

Note vs. the #60 thread: you floated chain-order (reverse walk) or a single preferred account. I went with the rate metric at my operator's request — it subsumes "one preferred account" (the highest-rate account *is* preferred, dynamically, and it generalizes to N without a per-account flag). The honest cost is that it's a computed metric rather than the order you set; happy to add a chain-order mode as an alternative if you'd prefer that be the default.

## agent

- Claude Code (claude-opus-5), operated by @alidarvishi14
- ran: `cargo fmt --all -- --check` (clean); `cargo clippy --all-targets --all-features --release -- -D warnings` (clean); `cargo test --release` (1704 passed, 0 failed, 2 ignored). operator verification: unknown — the operator directed the algorithm design but has not line-reviewed the diff.
- unsure: whether you'll want the rate metric vs. the chain-order walk you suggested (I can offer both, gated by a mode); the 0.8 hysteresis constant; leaving `scan_session_switches` (per-session decisions) on the old exhaustion-only path — this is the global-active decision only.


alidarvishi14:
Force-pushed one refinement (amended into the same commit): handle an **imminent weekly reset** in `greedy_burn_rate`.

The rate is `weekly_headroom / hours_until_7d_reset`, so as a weekly reset approaches `hours → 0` and the rate spikes toward infinity — which would yank every live session onto an account for the last minutes before its weekly refreshes, then off again once it resets. Now, within `GREEDY_MIN_HORIZON_HOURS` (5h ≈ a 5h window) of the reset, the account is valued as the fresh week it's about to become (`weekly_line / 168h`) instead of `remaining / near-zero`. That's continuous across the reset — no spike, no churn — and it's inert away from the boundary, so the normal water-filling behavior is unchanged. Added a unit test (`greedy_values_an_imminent_weekly_reset_as_fresh`).

`cargo fmt --check`, `clippy --release -D warnings`, and `cargo test --release` (1705 passed) all green locally.

uwuclxdy (changes requested):
Thanks for building this out, and for asking in #60 before you did rather than after.

Answering the question directly first, so you don't spend a round fixing things on an axis I'm dropping.

## direction

Going with the preferred-account option. Not because the rate metric is wrong. I checked the premise it rests on and it holds: `seven_day.resets_at` is one fixed absolute instant per account, so "hours until reset" is a real countdown and your denominator means what you say it means. Time-to-reset is genuinely something chain order misses, and you argued that well.

It's the predictability I want, and there's a second reason that only shows up once the pass runs per-session. More on that below.

Pulled the branch and ran it: `1725 passed, 0 failed` on debug, `clippy --all-targets --all-features -- -D warnings` clean. Didn't run `fmt --check`, `deny`, or `audit`, I'll take your word on those.

## what to build

Concretely, so there's no second guessing round:

1. **A per-account `preferred` toggle on the Fallback card, exclusive across the chain.** No Config-tab row at all. Marking a member is the opt-in, exactly like `last resort` needs no global arm. `toggle_last_resort` in `src/tui/app.rs:4765-4805` is the pattern to copy, including the save-failure rollback and the `moved_from` toast.
2. **`preferred` and `last_resort` are mutually exclusive**, each mark clears the other. A member carrying both is incoherent: the return would chase it, the sink rules forbid migrating live work onto it.
3. **Return only when preferred is `clear` and the active isn't itself preferred.** No dwell timer, no hysteresis constant. It's self-limiting: if preferred were clear, the exhaustion walk could only have skipped it when preferred was the spent one. That argument wants a test, not a comment.
4. **Freshness is a gate on both sides.** The return fires only when preferred's read is `decision_fresh`, and a stuck-`RateLimited` active never triggers one. Reasoning in the next section.
5. **Sessions return too, no cap.** Deliberate, see below.
6. **Both twins carry the pass**, plus wording that distinguishes a return from an exhaustion switch on the toast and the Overview hint.
7. **`preferred` wins wrap-off recovery.** If it's among the recovered members, `find_recovered_member` re-arms onto it instead of first-recovered. Otherwise you land on a sibling and return one tick later.

## what carries over regardless of the metric

These are properties of "switch toward a healthy target". They'd bite the preferred-account version identically, so they're the part of this review worth keeping.

**The pass already runs per-session, not just on the global active.** Your note says it leaves `scan_session_switches` on the old path. It doesn't: `build_chain_snapshot` is the shared body of both `snapshot_chain` and `snapshot_session_chain`, so `greedy` rides into the per-session leg and `scan_session_switches` calls the same `next_auto_switch_target`. Worth knowing because it changes what the feature is.

That's the second reason for preferring a named account. `walk_chain` starts at each session's own index and takes first-clear, so sessions diverge by construction. Any argmax is a global answer, so every live session resolves to the same member in the same tick and piles onto it 90s before the next usage fetch corrects anyone. A named home account has the same herding shape, and I'm accepting it there because "everything goes home" is the whole point of the feature. Accepting it in service of load balancing would have been self-defeating.

**`--with-fallback`'s help text is the contract and it's now false.** `src/cli.rs:254` says the session "swaps onto the next chain member when its window is spent". Whatever ships needs that copy moved in the same PR.

**Freshness.** The exhaustion walk treats a target's freshness as a preference rather than a gate, and the doc on `ChainSnapshot::fresh` says exactly why: so an exhausted active never loses its escape. A healthy active has no escape to preserve, so that exemption doesn't transfer. Right now a `Cached` member that hydrated at 7d 5% with a live reset three days out, actually sitting at 92%, outscores the fresh active and pulls every session onto an account one point from hard-blocking. `clear` reads the same stale entry, so nothing downstream catches it.

Same argument on the active side. `src/usage/scheduler.rs:2639` and `:2658` both promise that a stuck-`RateLimited` active with real headroom stays put, and that only a genuinely spent one moves. Such an active passes `reading_is_actionable`, isn't broken or canceled or kick-rejected, isn't exhausted, so it lands in the healthy early-return and gets moved on numbers the code has just declared untrustworthy. Gating both sides keeps those two comments true without editing them.

## on the current diff

Not asking you to fix these on this branch, but they'll recur:

- **The 5h clamp is a step, not a smoothing.** The comment calls it continuous across the reset boundary. It's continuous at `hours to 0`, discontinuous at `hours = 5` by a factor of `168 * remaining / (5 * line)`. Worked example: `a` active at 7d 74% resetting in 24h reads 1.0/h, `b` at 7d 50% resetting in 5.2h reads 9.23/h, so the chain moves to `b`. Twelve minutes later `b` crosses the 5h mark, its rate collapses to `98/168 = 0.583`, `a` reads 1.008, and the chain moves straight back. Two chain-wide relinks with zero usage change on either account, once per account per week.
- **`seven_day.utilization` arrives at 1-point integer resolution**, so `remaining` is quantized. A multiplicative band protects least where remaining is smallest: active at 96%, rival at 95%, both 24h out gives `0.083 < 0.8 * 0.125` and the whole chain moves on one integer point whose true value sits anywhere in 0.01 to 1.99. Any band near the line wants an absolute floor under it.
- **Every new test uses a 2-member chain**, so exactly one candidate ever reaches `best` and the `None => true` arm always fires first. Flip `rate > best_rate` to `<` and the suite stays green. The headline claim has no test. Same for `skip(i)`: no test sets broken, kick-rejected or canceled, and `clear` doesn't cover those axes, so deleting it lets the pass migrate onto an auth-broken account with everything still passing.
- **The row should be a `toggle_row`.** `src/tui/render/global_config.rs:707` spells out the rule: a pure on/off boolean is a toggle, not a 2-option cycle, since `on`/`off` in brackets read as a cycle. `refresh spent` is the precedent in the same file. Every other 2-value row uses named modes instead: `static`/`burn-aware`, `lazy`/`preemptive`, `stay on active`/`switch off all`. Moot if the Config row goes away, worth having in your head anyway.
- **The off-hint collides with the row two above it.** It reads "only switch off an account once it's spent", where `quota spent` in the same band uses "switch off" to mean signing every account out. Two adjacent rows, same words, opposite meanings.
- **Test comment arithmetic.** The rate comments are all stated in %/day; the code divides by hours, so the quoted figures are 24x off. Ratios survive so the asserts hold. `greedy_values_an_imminent_weekly_reset_as_fresh` quotes the weekly line as 90 in one clause and 97 in another. The header comment above the block says 98, which is the real value.
- Small one: the PR body says it reuses the walk's `skip`/`clear`/`fresh` closures. First two yes. The walk's `is_fresh` is declared after the early-return block, so you couldn't reach it and ended up with a second copy. Hoisting the existing one above the early return removes that.

The code itself is careful. Saying so plainly: no panic paths, lock discipline preserved, serde additive, both `ChainSnapshot` literal sites updated, `last_resort` excluded on both sides, third-party members going inert rather than mis-scored. The comments argue their decisions instead of narrating the diff, which is rarer than it should be.

Happy to take the preferred-account version as a fresh PR against this one or as a rewrite in place, whichever you'd rather. lmk if any of the seven above reads wrong to you, some of it is my read of what I want rather than something the code forces.

alidarvishi14:
Closing in favour of #62, which implements the preferred-account return design from your review here — fresh branch off `mommy`, no burn-rate metric, freshness gated on both sides, and non-vacuous 3-member-chain tests. Thanks for the detailed review; it shaped the whole rewrite.

PR #63, opened 2026-08-09
HTTP REST API

edgan:
It uses lego for LetsEncrypt SSL certificates.

It expects `/etc/lego/certificates` for `macOS/Linux`  and `lego\certificates` on `Windows`.

It uses the `FQDN` of the host for the SSL certificate.


uwuclxdy (changes requested):
Thanks for this, it's a lot further than I expected a first pass to get.

The reuse is right in the places I was braced to have to rewrite. `GET /v1/status` passes the `status.json` bytes through instead of building a second serializer, `POST /v1/switch` is a thin wrapper over `switch_profile_noninteractive` so it inherits the AUTH-1 gate and the divergence policy rather than re-deriving weaker ones, both new files go through `atomic_write_600`, and you extended `RankedMutex` instead of reaching for a raw `Mutex`. The framing discipline in `http.rs` is real work too: `Content-Length` only, chunked refused, disagreeing duplicates a hard error, framing errors close instead of resynchronizing.

Putting the listener on `clauth daemon` rather than a separate `clauth serve` is the right call and I'm keeping it. It makes the singleton question disappear instead of answering it.

## Blockers

### 1. `ApiSwitch` is held across `Rotation`

`routes::switch` takes the gate, then `switch_profile_noninteractive` reaches `ensure_installable` (`actions.rs:292`), which calls `RotationGuard::acquire` (`oauth.rs:1591`), which enters `rank::Rotation` at `runtime.rs:901`. `Rotation` is 100 and `ApiSwitch` is 380, and the check is strictly-greater, so it inverts. I reproduced it by minting a switch target with a clock-expired access token:

```
thread 'daemon::api::routes::tests::repro_...' panicked at src/lockorder.rs:206:17:
lock-order violation: acquiring rank 100 while holding [20, 380]
```

`Rotation` is missing from the derivation in the rank's own doc comment, which lists only `Config` and `State` beneath it. Nothing takes `ApiSwitch` while holding `Rotation`, so there's no cycle today and release builds are unaffected, but that assert is the whole enforcement mechanism and it's now firing on a false premise.

Every test in `daemon_api_routes.rs` mints `expires_at: None`, which is what keeps the suite off this branch: `expiring()` is false, so `ensure_installable` short-circuits before the guard. One case with an expiry catches it.

### 2. a poisoned switch gate is permanent

`RankedMutex::try_lock` collapses `Poisoned` and `WouldBlock` into one opaque `TryLockError`, and `routes::switch` maps that to `409 switch_in_progress`, which `wiki/Daemon.md` documents as "retry shortly". A poisoned mutex never clears, so after any panic inside the gate every later switch answers "busy" with nothing in flight, for the daemon's whole life. `RankedMutex::lock` deliberately recovers via `poison.into_inner()`; `try_lock` should too, or the poisoned case needs its own status and a log line.

Reachable from blocker 1 on a debug build, and independently from the `expect("config mutex poisoned")` in `actions.rs`.

### 3. `--rotate-token` doesn't revoke against a running daemon

`AuthToken::from_plaintext(&token::load_or_create()?)` runs once in `api::spawn` and nothing re-reads the file, so after `clauth daemon --rotate-token` the live daemon keeps accepting the **old** token and 401s the new one until restart. `--help`, `wiki/Daemon.md`, and `SECURITY.md` all say the opposite. `--rotate-token` conflicts with both `--replace` and `--listen`, so there's no single command that rotates and restarts.

This is the response to a leaked bearer token, so it's the one control that has to work.

### 4. `println!` in `src/main.rs:169,172`

Fails `out::tests::no_bare_print_macro_under_src`. `src/out.rs` owns stdout for the reason the guard test exists: `clauth daemon --print-token | head -1` panics and exits 101 instead of exiting 0. `outln!` is the fix.

### 5. `cargo fmt --all -- --check` is red

Four hunks: `src/daemon/api/tls.rs:211`, `tests/inline/cli.rs:748`, `tests/inline/daemon_api_tls.rs:218` and `:233`. CI runs it as its own leg.

### 6. `--replace --listen` can leave no daemon at all

`api::spawn` runs after the singleton claim, so under `--replace` the incumbent is already SIGTERM'd and reaped by the time the certificate is read. A missing or mismatched cert then propagates out of `serve` and the box has no daemon, so refresh and auto-switch stop, not just the listener. `wiki/Daemon.md` recommends `clauth daemon --replace --listen` as the post-`lego renew` hook, so the documented automation is the trigger. Loading the cert and binding the socket before the claim makes a bad renewal a no-op instead of an outage.

## Design calls

These are mine to make, and three of them go against what you built. Reasoning inline so you can push back on any of them.

**The bind default moves to loopback.** Bare `--listen` binds `127.0.0.1:8443`; an explicit `--listen 0.0.0.0:8443` is accepted exactly as typed, no second flag. I take your point that a listener the phone can't reach is a useless default for this flag, and the flag still has to be passed either way. It's the shortest-spelling-is-widest-exposure part I don't want on a tool whose job is credential custody. `tests/inline/cli.rs:748` pins the current behavior and moves with it.

**`--cert`/`--key` alongside lego.** Keep the lego path as the zero-config default, it's genuinely nice. But as it stands the only startable deployment is a host with a publicly resolvable FQDN plus a Let's Encrypt cert, and the mode I expect most people to use is a tailnet. `tailscale cert` writes `<machine>.<tailnet>.ts.net.crt` and `.key`, no issuer file, in a directory `hostname -f` won't name. `server_config_from` already takes explicit paths, so this is thin.

**`/v1/*` becomes `/api/v1/*`.** A string change now. There are two clients coming that generate their models from the OpenAPI spec, so after that it's a version bump for nothing.

**`auth_token.json` gets a capability tier field now.** The single global token is proportionate while switch is the only mutation, so it stays. But view-only vs control per device is coming, and adding a tier to an existing single-token file later is a migration rather than an addition. A `"tier": "control"` written today makes it additive.

## Smaller things

- `?all=1` passes `None` for `LiveSignals` where the daemon's own writer passes `Some(&live)` (`daemon/mod.rs:657`), so on one live daemon `/v1/status` and `/v1/status?all=1` disagree on `fetch_status`, `next_refresh_at`, `stale` and `pending_switch`. The wiki says `?all=1` only adds disabled accounts. `ApiContext` can't do better as constructed, it holds no scheduler stores.
- `SECURITY.md` and `wiki/Daemon.md` both say an unauthenticated request can't occupy a connection slot. `claim_slot` fires at `accept()`, before the handshake and before any token check, so 32 silent TCP sockets lock out every real client with no token and no TLS. The 401-closes rule is real, it just isn't what that sentence claims.
- The nine end-to-end TLS tests shell out to `openssl` and return green with a `SKIPPED` line when it isn't usable. `-addext` isn't supported by macOS's LibreSSL and Windows runners have no guaranteed `openssl`, so two of the three CI legs can carry zero listener coverage while reporting pass. A vendored fixture cert, or failing instead of skipping, would close it.
- `CLAUTH_NO_API` has no test. `tests/inline/update.rs:166` pins `updates_enabled()` for `=1`, `=0` and `=true`, which is the shape to copy.
- `tests/inline/feature_coverage.rs` maps each README `## Features` bullet to test-name prefixes, and its doc comment asks for a prefix when a bullet starts claiming something new. The "Headless" bullet now claims the REST API with no prefix added.
- `tls.json` doesn't get the 0600 pin that `auth_token.json` got. Same helper, same invariant, it applies to every clauth-owned file rather than just the secret ones.
- `Content-Length` goes through `v.trim().parse::<usize>()`, which accepts `+5`, `\t5`, and non-breaking-space forms that httparse permits and nginx or haproxy reject. Only exploitable behind a proxy, which isn't the documented deployment, but it's the one gap in the framing story the module doc says is closed. Requiring all `b'0'..=b'9'` is a line.
- `ConnectionSlot::drop` clears `SATURATION_LOGGED` unconditionally, so the "at capacity" line fires once per slot release rather than once per episode. A connect-and-close loop against a full server gets one log line per refusal, which is what the flag exists to prevent.
- A truncated `auth_token.json` is silently replaced with a fresh token and no log line, while the schema-newer branch does log. The remote client 401s with nothing in `daemon.log` naming the cause.
- `wiki/Daemon.md` advertises a 10s deadline per read or write without saying a `POST /v1/switch` can legitimately take ~25s on the state flock, or ~20s on a macOS Keychain shell-out. A client sizing its own timeout off the 10s aborts mid-switch.
- `claim_slot` is load-then-`fetch_add` with the cap held by a comment about there being one accept thread. `fetch_update` makes it structural for the same cost.
- `src/daemon/api/tls.rs:11` has your real hostname in the module doc's example paths. `wiki/Daemon.md` uses `<fqdn>` throughout, worth matching.

## What I ran

At `a59b8b6`, in a worktree with its own `CARGO_TARGET_DIR`:

| | |
|---|---|
| `cargo clippy --all-targets --all-features --release -- -D warnings` | pass |
| `cargo fmt --all -- --check` | **fail**, 4 hunks |
| `cargo nextest run --release --all-targets --all-features` | **fail**, 1868/1869 |
| `cargo nextest run --all-targets --all-features` (debug) | **fail**, same one test |
| `cargo deny check` | pass |
| `cargo audit` | pass, one pre-existing allowed advisory |

Your base is `e607814`, which is `mommy`'s tip, so none of that is inherited. I didn't run the openssl-gated tests on macOS or Windows, and I didn't test `--replace --listen` rebinding on Windows, where `std` doesn't set `SO_REUSEADDR`.

## On my side

Some of what's above wasn't visible to you. The wiki rows are the clearest case: `wiki/Quickstart.md` owns the command table and its `clauth daemon` row still lists four flags, `wiki/Security.md`'s env-var table is missing `CLAUTH_NO_API`, and `wiki/Home.md`'s Daemon row got left behind while the README's equivalent row moved. The rule that Quickstart owns that table lives in a file that isn't in the repo, so that one's on me.

The brief I sent you was wrong about one thing too: it says `docs/daemon.md` asserts the daemon has no external mutation surface and no socket. It doesn't. Only `wiki/Daemon.md` carries that stance, and you rewrote it correctly.

Two conventions from `.github/CONTRIBUTING.md` for the next push: code and docs go in separate commits, and both the commit header and the PR title follow Conventional Commits, lowercase and imperative. `Added daemon --listen, ...` is 73 characters, past tense, and carries no type or scope.

edgan:
> The bind default moves to loopback. Bare --listen binds 127.0.0.1:8443; an explicit --listen 0.0.0.0:8443 is accepted exactly as typed, no second flag. I take your point that a listener the phone can't reach is a useless default for this flag, and the flag still has to be passed either way. It's the shortest-spelling-is-widest-exposure part I don't want on a tool whose job is credential custody. tests/inline/cli.rs:748 pins the current behavior and moves with it.

I understand the `paranoia` , but here are a few thoughts.

1. `--listen` is not the default. `clauth daemon` doesn't bind. `clauth daemon --listen` does. In addition the default could be declared in help.
2. Default binding to `127.0.0.1:8443` is useless, and it would be simpler to just require the user to specify than even making it the default.
3. If you are exposing this to the internet, you made a choice.
4. It still requires the `auth token`, and it uses SSL. Personally this is good enough for me.
5. All the major operating systems and distributions come with firewalls on by default.

edgan:
> Added daemon --listen, ... is 73 characters, past tense, and carries no type or scope.

IMHO you should change your `.github/CONTRIBUTING.md` on the past tense rule. It has to be the single dumbest rule I have ever heard about git usage. The git log is a log. Logs are about things that already happened. I don't `Place the apple in the cupboard`. Instead I `Placed the apple in the cupboard`. Most projects I see write commit messages in `English`, so use good `English` grammar.

I personally don't follow the 72 character `GitHub` rule. I know it cuts them or `...` them. There are some many broken things about `GitHub`. As for [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), this is a open source project, not a corporate piece of software. The whole point of those hyperactive commit tagging rules is different forms of automation. But this is a small project that isn't moving in that direction.

Also being overly prescriptive on the rules of any contribution is a recipe for having no contributors. I have to pull teeth to even get some of my users to file `GitHub` issues. They all want to ask questions or report their issue on `Reddit` or `Discord`.

If you want to be strict about all this I will tell you to take over the PR, and do the work yourself.

uwuclxdy:
1-3 and 5. Agree.
4. The token gates the routes, not the port. Slots are claimed at  `accept()`, before the handshake and before the token is read, so 32 silent connections take every slot for 10s each. Reconnecting keeps it that way.

Commits and PR titles dropped from `.github/CONTRIBUTING.md` since the repo squash merges anyway.

Everything else in the review stands. Slices 6 through 12 are mine to do.

uwuclxdy:
Round 2, read at `07d9a752`.

The four blockers are genuinely fixed, and three of them are pinned by a test that reds on revert: the clock-expired lock-order case, the poison recovery with its negative control, and `rotating_the_token_revokes_the_old_one_against_a_live_context`. `/api/v1` behind one constant, the `tier` field, `--cert`/`--key`, `CLAUTH_REQUIRE_TLS_FIXTURE` and the `?all=1` live-signals fix all landed. `LiveStores` is a better answer to `?all=1` than the one I asked for.

Three things are broken at that head, plus a set of smaller ones. I'm landing all of it on a branch off your head rather than sending you another round.

## 1. The bind moved above the singleton claim

`api::prepare` reads the certificate and calls `TcpListener::bind`, and `daemon::serve` runs it before the claim. The certificate half is what round 1 asked for and it is right. The bind half is ordered against itself: under `--replace` the port's holder is the incumbent, and releasing it is the claim's own job.

Measured here on Linux, with `SO_REUSEADDR` on both sockets, which is what `std::net::TcpListener::bind` sets on unix:

```
incumbent: TcpListener::bind(0.0.0.0:45881) OK
replacement(same wildcard): ERR kind=AddrInUse raw_os_error=Some(98)
replacement(narrowed to 127.0.0.1): ERR kind=AddrInUse raw_os_error=Some(98)
replacement(after incumbent released): OK
```

`SO_REUSEADDR` governs `TIME_WAIT`, not `LISTEN`; `SO_REUSEPORT` is the option that would allow the overlap and nothing sets it. So:

- `clauth daemon --replace --listen` against a listening incumbent dies at the bind. The incumbent is never signalled, never reaped, and keeps serving the certificate it read at boot, which is the state `wiki/Daemon.md` names as the failure mode to watch for. `--replace --listen` is the post-`lego renew` hook that same page recommends.
- `clauth daemon --standby --listen` never reaches `stand_by`. It exits non-zero, and under `KeepAlive{SuccessfulExit=false}` the supervisor restart-loops while the standby slot stays empty.
- plain `clauth daemon --listen` with a daemon already up exits 1 on "address already in use", where `ExitIfRunning`'s own doc contracts exit 0 with `already running (pid N); exiting`.
- a standby that can bind parks holding a `LISTEN` socket nothing accepts on. A client `connect()` succeeds and then hangs, where an unbound port answers `Connection refused`.

The split you made, one step further: certificate above the claim, socket below it, and below promotion for standby.

## 2. The TLS fixture is dead code on Windows

`SERVER_NAME`, `openssl`, `REQUIRE_TLS_FIXTURE_ENV`, `no_tls_fixture` and `generate_chain` moved into `src/testutil.rs`, and `tests/inline/daemon_api_server.rs`, which is `#![cfg(unix)]`, is still their only caller. CI runs `cargo clippy --locked --all-targets --all-features --release -- -D warnings` on `windows-latest`. Reproduced with the `x86_64-pc-windows-gnu` cross-lint:

```
error: constant `SERVER_NAME` is never used
error: function `openssl` is never used
error: constant `REQUIRE_TLS_FIXTURE_ENV` is never used
error: function `no_tls_fixture` is never used
error: function `generate_chain` is never used
error: could not compile `clauth` (bin "clauth" test) due to 5 previous errors
```

The comment above the block says the fixture is shared with the proxy's outage test. There is no proxy test in this repo, so moving the block back under the `cfg(unix)` module closes both.

## 3. `CLAUTH_REQUIRE_TLS_FIXTURE` only catches openssl being absent

`generate_chain` asserts behind `if !openssl(&["version"])`. Every later failure returns `None`, and each of the ten call sites turns that into `eprintln!("SKIPPED ...")` and a pass. macOS LibreSSL rejecting `-addext` is exactly the present-and-failing case, so the leg CI now marks required can still carry zero listener coverage. Calling `no_tls_fixture()` on the generation-failure arm too is the one-line version.

## Smaller

- `GET /api/v1/status?wait=0` with an `If-None-Match` returns `304` without ever reading `status.json`: `deadline` is `now`, and the loop checks it before its first read. A client computing `wait` as a remaining budget is told nothing changed, indefinitely. The same edit removes the 250ms first-read delay at `wait>0`.
- `wait_for_status_change` skips only a read `Err`, and `etag_for` digests the raw bytes when the body does not parse, so a half-written feed is answered `200` with the truncated body and a fabricated tag. Production writes atomically so this needs a hand-edited file, but the comment above it claims mid-replacement is waited through.
- the per-request `logline!` fires for the 401 too, and `request.path` is bounded only by `MAX_HEAD_BYTES`, so an unauthenticated peer writes ~8 KB of chosen bytes into `daemon.log` per connection.
- `token::current_or` does `read_valid().ok().flatten()`, so the unknown-`tier` error that `load_or_create` refuses to start on is silently ignored on the live path and the daemon keeps its spawn-time token. A newer clauth's `--rotate-token` writing an unknown tier reinstates the bug blocker 3 fixed.
- `Response::refused`'s `reason` is closed on the `AuthGate` arms, and the IO arms below them propagate `anyhow` context naming absolute paths under `$HOME`.
- `read_valid`'s doc says it is "quiet on purpose" because `current_or` calls it once per request, and three lines below it logs the schema-too-new line unconditionally. A daemon downgraded onto a schema-2 file writes that line once per request.
- the four `actions.rs` republish sites run under the caller's config mutex on the CLI, TUI and MCP paths, doing the profile-cache and flock sweep the daemon's own writer deliberately keeps outside that lock. They fire when the switch was a no-op too.
- a non-daemon publish stamps a fresh `generated_at` with `live: None`, and `wiki/Daemon.md` tells readers a fresh stamp means a live daemon, and to poll the feed rather than the CLI when the fetch outcome matters. The republish is worth keeping, it just has to stay distinguishable.
- `wiki/Daemon.md:245` says to treat `409` as the retryable answer. `routes.rs` answers `503 state_locked` for the flock and `409` for the refusals that need a human, and the route table two paragraphs up has it right.
- the refusal shape promises `reason` on five failures. `404 profile_not_found`, `409 switch_in_progress` and `400 bad_request` go through `Response::error`, which emits none.
- `wiki/Daemon.md:348` still calls `clauth status --json --all` the one way to reveal disabled accounts, which `?all=1` falsified in this PR.
- `wiki/Configuration.md`'s storage tree closes with "That is the whole tree: every path clauth writes is listed above" and lists neither `auth_token.json` nor `tls.json`. Following it deletes the token file and revokes every client.
- four doc comments describe a `/api/v1/mirror` route and a `MirrorBody` type that are not here (`routes.rs:1,216`, `http.rs:42`, `token.rs:31,34`), and `token.rs:177`'s traffic model rests on the same replica.
- the extraction left three doc comments on the wrong item: `Daemon`'s doc now sits on `LiveStores`, `write_status`'s on `live_stores`, and `serve`'s on `struct Prepared`, each leaving the real item undocumented.
- `the_advertised_budget_is_the_one_enforced` sets `lifetime: 700ms`, so `as_secs()` is `0` and the assertion reduces to `advertised == 0`. `max=` is read by no test in the tree.
- `a_wait_is_not_woken_by_the_timestamp_alone` failed once in four full debug runs here and passed 10 of 10 alone. Its ticker writes the feed with `std::fs::write` where production goes through `atomic_write_600`, so the reader can catch a half-written file.
- the ordering fix in 1 has no test. `a_missing_certificate_fails_in_prepare_not_after_the_claim` calls `prepare` directly, and the suite's only `serve` caller passes `listen: None`.
- `CLAUTH_NO_API` is pinned at the predicate and not at the call site, so deleting the `if api_enabled()` guard leaves the suite green. `update.rs` pins both.
- `HEAD /api/v1/health` answers `405`, and `method.to_ascii_uppercase()` accepts a lowercase verb.
- stale `/v1/` in `src/daemon/mod.rs:418,420`, and the `lockorder` module doc still names `RotationGuard` as outermost with `ApiSwitch` absent from its derivation list.

Gate at `07d9a752`, own target dir: `check --locked`, clippy, `fmt --check`, `deny` and `audit` all pass. Release `nextest` 3139/3139. Debug `nextest` 3143/3143 on three runs and 3142/3143 on a fourth, the flake above. The windows cross-lint is the only hard red.

PR #64, opened 2026-08-11
fix(credentials): preserve mcpOAuth blocks across a profile switch

theo-nash:
> **operator asked:** "can't we just update the clauth code to copy over the mcp oauth?" — make account switching stop wiping MCP-server logins. The ask evolved: first run the patched build locally, then (once it was working) open this upstream because it's a real bug others hit.

clauth's account switch drops every `mcpOAuth` block from `~/.claude/.credentials.json`, silently logging you out of all MCP servers on each switch. `ClaudeCredentials` models only `claudeAiOauth`, so the snapshot-then-symlink switch installs a login-only store and the per-MCP-server logins — which are independent of the Claude account — are lost. Adoption ("new from current profile" onto an existing untracked login) also hard-refused, because the switch guard byte-compared the whole live file against the login-only store and tripped on the mcpOAuth difference.

This preserves non-login top-level blocks without widening the typed model:

- `save_profile` re-attaches any non-login keys already in the store, so a login-token refresh no longer drops them.
- switching (`link_` / `force_link_profile_credentials`) carries the live file's non-login blocks onto the incoming profile before the swap.
- the `link_profile_credentials` refuse-guard compares the login access token (matching what `classify_credentials_link` already does) instead of the whole file, so a live file that differs only in mcpOAuth is no longer a false refusal; a genuinely different login is still refused.

The login (`claudeAiOauth`) is never altered, and the blocks preserved are per-MCP-server logins that don't belong to any one Claude account. I deliberately did not add a field to `ClaudeCredentials`, so its ~35 constructors are untouched. I didn't find docs describing this behavior, so there was nothing to update.

## agent

- Claude Code (Claude Opus 4.8), operated by @theo-nash
- ran: `cargo test` → 1786 pass, 1 fail (`reload_fingerprint_ignores_a_bare_session_token_stamp`) which also fails on unmodified `mommy` — a filesystem mtime-resolution flake on my WSL2 box, not from this change; `cargo fmt --check` clean; `cargo clippy --all-targets` no warnings in the changed files. Added 6 tests (synthetic tokens only): the carry helper, save-refresh preservation, an end-to-end switch, the adoption link, and the retained different-login refusal. operator verified: ran the patched build on a real two-account setup — an actual account switch preserved all 12 of their mcpOAuth blocks.
- unsure: narrowing the `link_profile_credentials` refuse-guard from a whole-file compare to a login-token compare is the one change to a safety guard — it matches `classify_credentials_link`'s existing access-token comparison and the `switch_replaces_active_account_mirror_without_refusing` test still passes, but it's worth a maintainer's eye. Separately, I chose a localized JSON merge over adding a flattened `extra` field to `ClaudeCredentials`; the typed approach may fit the codebase better but would churn ~35 constructors.

### DIFF
--- a/src/claude.rs
+++ b/src/claude.rs
@@ -432,6 +432,48 @@ pub(crate) fn create_symlink(target: &Path, link: &Path) -> Result<()> {
         .context("failed to copy credentials")
 }
 
+/// Carry the live credential's non-login blocks — notably `mcpOAuth`, the
+/// per-MCP-server logins Claude Code keeps alongside the Claude account — onto
+/// `target` before it becomes the live credential. These blocks are independent
+/// of which Claude account is active, so an account switch must preserve them:
+/// without this, switching to a profile whose store lacks them (e.g. a fresh
+/// browser login) drops every MCP-server auth. Each non-login top-level block
+/// present in the live file is copied onto the target (its prior copy of that
+/// block overwritten); a block the live file lacks is left as the target has it.
+/// The target's `claudeAiOauth` login is never touched. A no-op when the live
+/// file is unreadable, carries no such blocks, or `target` is not an OAuth store
+/// (a session-token sidecar has no `claudeAiOauth`).
+fn carry_live_extra_into(link: &Path, target: &Path) -> Result<()> {
+    let Ok(live) = read_json_file::<serde_json::Value>(link) else {
+        return Ok(());
+    };
+    let Ok(mut stored) = read_json_file::<serde_json::Value>(target) else {
+        return Ok(());
+    };
+    let (Some(live_obj), Some(stored_obj)) = (live.as_object(), stored.as_object_mut()) else {
+        return Ok(());
+    };
+    // Only an OAuth store carries these blocks; never touch a session-token sidecar.
+    if !stored_obj.contains_key("claudeAiOauth") {
+        return Ok(());
+    }
+    let mut changed = false;
+    for (key, value) in live_obj {
+        if key == "claudeAiOauth" {
+            continue;
+        }
+        if stored_obj.get(key) != Some(value) {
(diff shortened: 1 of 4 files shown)


uwuclxdy:
Merging this, thanks for the writeup. The premise is right and it closes something that's been sitting in my backlog: whether to teach the link path to merge unknown top-level keys forward, or accept the loss as consistent with what every switch already did. You picked the branch I wanted.

Answering the two things you flagged.

Narrowing the refuse-guard is the right call. The whole-file compare is the original formulation from when that guard first went in, not something later hardened into place, so replacing it isn't undoing an argued decision. There is one state it now lets through though. When both sides read `None` the compare passes and the live file gets removed with nothing to relink. `live_login` is `None` for a torn mid-write file or for one carrying no `claudeAiOauth`; `target_login` is `None` whenever the install source doesn't exist, which is every api-key profile. Daemon boot and TUI start both call `link_profile_credentials` on the active profile and discard the result, so it's silent on those two. Keeping the byte compare as the fallback when either side yields no token closes it.

The localized merge over a flattened `extra` field is right too. Agreed on not churning the constructors for this.

Two changes I'll layer on top. The carry becomes an allowlist of `mcpOAuth` rather than every key that isn't `claudeAiOauth`, since that store is a single object Claude Code mutates wholesale and I'd rather whatever lands in it next didn't cross accounts silently. And the carry goes best-effort instead of `?`-fatal, so an unwritable profile dir can't strand a switch.

The rest of the followups are on me.

theo-nash:
Thanks for getting this merged.  I needed it for clauth to work for me so happy to share.

uwuclxdy:
yeah, i can imagine how annoying it used to be before lol thanks for the pr

Issue #17, opened 2026-07-05
Bug: Fallback accounts fails


edgan:
I have three accounts, and I have seen the faillback fail. It does successfully replace the symlink, but /status after the switch or closing `Claude Code` and then /status show it still thinks it is on the same account. Just replacing `~/.claude/.credentials.json` doesn't seem to be enough. When it is broken the solution is to logout and back into an account that isn't 5h limited.

My auto switch is set to `100%` on all three accounts.

`Claude Code` version: `2.1.201`
`clauth` version: `master` aka `71ce018d7ad516f61dd0765181141a5018b7a605`

uwuclxdy:
Thanks for the bug report, this might be the root cause:

Every clauth switch (manual or fallback, same code path) repoints `~/.claude/.credentials.json` and rewrites settings.json. It doesn't change the `oauthAccount` identity block cached in `~/.claude.json`, which claude code trusts on startup to show which account is active. This is purely cosmetical and the API calls will go to the account depending on the credentials file. There have been past efforts to address this, might be time to take another look now. `/login` fixes it just because a full oauth flow rewrites the block, which is the thing that clauth switch skips.

will have to: probe how CC 2.1.201 behaves when the block is absent (re-derive vs login prompt), that decides whether the fix is "drop the stale block on switch" or "snapshot + restore the block per profile". either way it lands in the shared switch path, so manual + fallback switches both get it.

uwuclxdy:
Fixed in `f3b754a`. Every switch (manual + fallback, same code path) now drops the stale `oauthAccount` block from `~/.claude.json` so CC re-derives identity from the live credentials.

PR #66, opened 2026-08-19
fix(keychain): pick the transport by size, never let security truncate

agustinprod:
## The defect

`security -i` reads one command per line into a bounded buffer. Past that buffer it does **not** refuse. It splits the line, runs the head with a **truncated** `-w` value, and reports the tail as `unknown command`. The item is left holding truncated JSON.

I hit this as a real data loss on macOS 15 / Darwin 25, on an account with ~20 OAuth MCP servers. The write meant to *preserve* `mcpOAuth` is the one that destroyed it, along with `claudeAiOauth` beside it, leaving the item unparseable:

```
Error: Keychain write failed (security exit 1): security: unknown command "CJuYW1lIjoi…"
```

That fragment is the middle of a JWT from the blob being written.

## Measurement

Throwaway service, same host:

| payload | `security -i` | argv |
|---|---|---|
| 3900 B | intact | intact |
| 4100 B | **BROKEN** | intact |
| 32000 B | **BROKEN** | intact |
| 64000 B | — | intact |
| 128000 B | — | **BROKEN** |

So `-i`'s ceiling is a 4096-byte command line, and argv holds to 64 KiB. argv's ceiling is not `ARG_MAX` — that is 1 MiB here — but `security`/Keychain's own.

## Why it appeared now

Unreachable while the mirror wrote a login-only blob: 1–2 KB never came near 4096. #64 changed the size class by preserving Claude Code's sibling keys, and a blob carrying `mcpOAuth` for a dozen-plus OAuth MCP servers clears the ceiling easily. #21 and #64 are each correct alone and collide together.

`run_with_deadline`'s comment reasons about the 64 KB pipe buffer, which is the wrong limit — the pipe is fine, the line tokenizer is not.

## The change

Choose the transport by size: `-i` while the line fits, argv above it, refuse past 64 KiB rather than write something `security` will mangle.

The argv fallback gives up the Endpoint-Security-log property #21 won, and only for blobs too large to keep it. argv stays same-UID-or-root only — the exposure #21 recorded as already accepted (TECH-9 #17). A truncated item is a lost login *and* lost MCP servers, so correctness takes precedence. Happy to gate the fallback behind a setting, or to move the whole write to `SecItemUpdate`, if you would rather not lose the property at all.

## Tests

- `stdin_carries_a_line_up_to_the_ceiling` / `argv_takes_over_up_to_its_own_ceiling` — pure, pins both ceilings and the refusal.
- **KC-3** `a_blob_past_the_stdin_ceiling_round_trips_intact` — drives a realistically sized blob through the real `security` on a throwaway service. **Fails on the parent commit, passes here.**

Full suite: 2449 passed, 4 ignored. `clippy --all-targets` clean.

### DIFF
--- a/src/keychain.rs
+++ b/src/keychain.rs
@@ -194,6 +194,55 @@ fn run_with_deadline(
 /// Keychain generic-password service Claude Code reads/writes for its login.
 const SERVICE: &str = "Claude Code-credentials";
 
+/// Longest command line `security -i`'s tokenizer reads as ONE command.
+///
+/// Measured on macOS 15 / Darwin 25 against a throwaway service: a 3900-byte
+/// value round-trips byte-identical, a 4100-byte value does NOT. Past the
+/// ceiling `security` does not refuse — it splits the line, executes the head
+/// with a TRUNCATED `-w` value, and reports the tail as `unknown command`. The
+/// item is left holding the truncated JSON, so a write that overruns this
+/// silently destroys whatever the item held.
+///
+/// This mattered nowhere until the mirror started preserving Claude Code's
+/// sibling keys: a login-only blob is 1-2 KB and never came close, while a blob
+/// carrying `mcpOAuth` for a dozen-plus OAuth MCP servers clears it easily.
+const SECURITY_STDIN_LINE_MAX: usize = 4096;
+
+/// Largest value that survives the argv transport intact, same measurement run:
+/// 64 KiB round-trips, 128 KiB does not. NOT an `ARG_MAX` limit — that is 1 MiB
+/// on this host — so the ceiling is `security`/Keychain's own. A blob past this
+/// is refused rather than written, because every transport we have would mangle
+/// it and a mangled item is a lost login plus lost MCP servers.
+const SECURITY_ARGV_VALUE_MAX: usize = 64 * 1024;
+
+/// How [`put_blob_at`] hands one write to `security`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PutTransport {
+    /// `security -i`, command line over stdin. Keeps the token out of this
+    /// process's argv, hence out of Endpoint Security exec logging (PR #21).
+    Stdin,
+    /// `security add-generic-password …` with the value as a real argv word.
+    /// Gives up the EDR-log property to stay under [`SECURITY_STDIN_LINE_MAX`];
+    /// argv is still same-UID-or-root only, the tradeoff PR #21 called already
+    /// accepted. Correctness wins: the alternative is a truncated item.
+    Argv,
+}
+
+/// Pick the transport for a `-i` command line of `line_len` bytes carrying a
+/// `value_len`-byte password. PURE, so the decision and both ceilings are pinned
+/// without a Keychain.
+fn put_transport(line_len: usize, value_len: usize) -> Result<PutTransport> {
+    if line_len <= SECURITY_STDIN_LINE_MAX {
+        return Ok(PutTransport::Stdin);
+    }
+    if value_len <= SECURITY_ARGV_VALUE_MAX {
+        return Ok(PutTransport::Argv);
+    }
+    anyhow::bail!(
+        "refusing to write a {value_len}-byte Keychain item: past the {SECURITY_ARGV_VALUE_MAX}-byte          ceiling every `security` transport truncates instead of failing, which would destroy the          login and any MCP server logins stored beside it"
+    )
+}
+
 /// `security(1)` exit status for `errSecItemNotFound` (-25300). Returned when no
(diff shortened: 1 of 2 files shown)


uwuclxdy:
The truncation mechanism is real, reproduced on macOS 26.5.2 (Darwin 25.5.0, arm64): a 4097-byte `security -i` line round-trips, a 4098-byte one truncates at 4096, exits 1, and leaves the item holding the truncated value. `line_len <= 4096` counting the trailing `\n` is safe as written, so the stdin side needs no change.

The argv ceiling doesn't reproduce. On that box `-w` values round-trip intact to 1,047,900 bytes (65536, 128000, and 1000000 all intact), and the break at ~1,048,000 bytes is exec E2BIG (`Argument list too long`, item untouched): `kern.argmax` minus argv overhead, not a `security`-internal limit. The 128000-byte row in the table above reads BROKEN where it's intact here.

Could you re-run the argv rows on your box and report back the OS and the rows? The table says "macOS 15 / Darwin 25", but Darwin 25 is macOS 26, so which version the measurement ran on is unclear, and if 128000 breaks there, something differs between the two hosts.

Changes this needs:

- Raise `SECURITY_ARGV_VALUE_MAX` from 64 KiB to 512 KiB. The current cap refuses writes that work, and the refusal text claims every transport truncates past it, which the measurement contradicts. 512 KiB leaves a 2x margin under the measured break, where E2BIG still fails safe with the item untouched. The constant's doc comment needs the same rewrite: the ceiling is the exec limit, not `security`'s own.
- Update `put_blob_at`'s doc comment. It still says the write goes over stdin "so the token never appears in this process's own argv", which the argv branch now falsifies for blobs past the line ceiling. The EDR property #21 won is deliberately dropped there, and the doc should say so.
- Fix the refusal string: `65536-byte          ceiling` and `destroy the          login` carry literal 10-space runs, and the message names no remediation.
- `security_error`'s parenthetical "travels only on the child's stdin" is false for the argv transport; the operative claim is that only stderr gets embedded.

The two new pure tests and KC-3 pass here; the full-suite count is taken as reported.

harrisrobin:
Third-party data point on the argv rows, since @uwuclxdy asked for another host.

I hit this bug independently as a user (repeated forced re-logins, an item left holding an empty `accessToken` and `expiresAt` 0) and traced it to the same 4096-byte `security -i` line cap before finding this PR.

**Host:** macOS 26.6.2, Darwin 25.6.0, arm64. `kern.argmax` = 1048576.

**stdin (`security -i`)** — line lengths including the trailing `\n`, throwaway service:

| line | result |
|---|---|
| 4090–4097 B | rc=0 |
| 4098 B and up | rc=1, `security: unknown command "xxx…"` |

Matches @uwuclxdy exactly: `line_len <= 4096` excluding the newline is the correct boundary.

**argv (`add-generic-password -U -w <value>`)**, value length, written and read back:

| payload | result |
|---|---|
| 3900 B | intact |
| 4100 B | intact |
| 32000 B | intact |
| 64000 B | intact |
| 65536 B | intact |
| 128000 B | intact |
| 262144 B | intact |
| 524288 B | intact |
| 1000000 B | intact |
| 1047000 B | E2BIG (`argument list too long`), item untouched |

So the 128000 B row in the PR table does **not** reproduce here either. This host agrees with @uwuclxdy: argv holds to ~1 MiB and the break is the exec argument limit failing safe, not a `security`-internal ceiling. `SECURITY_ARGV_VALUE_MAX` at 64 KiB would refuse writes that work on at least two hosts.

For what it's worth on "why now": my item was 4981 bytes with ~20 OAuth MCP servers, i.e. exactly the #21 + #64 collision described above. The failure mode is worse than a lost write — clauth then read back unparseable JSON (`EOF while parsing a string at line 1 column 3742`) and the recovery path left a blank login, so every new `CLAUDE_CONFIG_DIR` came up signed out.

Happy to re-run any row or a different shape if it helps close out the disagreement.

uwuclxdy:
Landed as `be5c39e2` with your implementation as the base, plus the changes from the review: the cap moved to 512 KiB (your 64 KiB ceiling and the 128000-broken row did not reproduce — argv is intact to 1,047,900 B and breaks at exec `E2BIG` with the item untouched, so the ceiling is the exec limit, not `security`'s own), the refusal and doc comments rewritten to that provenance, and the boundary pinned to the measured 4096 by a literal test so the constant cannot drift above it. #76 reported the same defect independently; its rotation-caller test and the argv disclosure event line are folded in.

Two follow-ups landed beside it, same class: `d787c3ed` verifies every write by reading it back and quarantines an unparseable item before the overwrite or delete (both #76 reporters lost their logins to that recovery path, not to the truncation itself), and `3b482282` drains the subprocess pipes concurrently so an item past the pipe buffer no longer deadlocks its own read.

Thanks for the report — the measurement table is what made the review fast.

PR #67, opened 2026-08-20
feat(usage): interleave the 5h warm-up kick, and show the ring

tmanaud:
> **operator asked:** "do you have everything to create a PR (check public repo how to propose changes first)" — the final form of an ask that evolved across the session: adversarial review of the auto-start design, empirical validation of its API assumptions, then filing the result.

This PR queues the `auto_start` 5h window-opening kicks across accounts so a
fresh window lands every `5h / N` instead of all windows resetting at once.

With several accounts on `auto_start`, every window reopens the instant it
lapses, so they stay in phase and reset together — everything at once, then
five hours of nothing. Opted-in queue members may open a window only when no
other member opened one within `5h / N` (less a capped poll-jitter tolerance).
One member is elected per tick before the fetch fan-out, so parallel workers
cannot double-open.

The queue anchor is derived from each member's existing
`usage_history.jsonl`, with no new state file. A single `/usage` snapshot cannot
distinguish an idle window from a just-opened one: both can report a reset one
full window ahead. Across two samples, however, an idle reset slides with the
clock while a real window keeps a fixed boundary. The newest boundary confirmed
by an agreeing pair becomes the anchor. The scheduler seeds that result into
memory at startup and advances it immediately when a lapsed-window kick lands.

The global `auto_start_queue` switch defaults off, so upgrades do not change
existing behaviour. Queue membership is one shared rule for the scheduler,
`status.json`, and the TUI: enabled, uncanceled, non-quarantined profiles that
opt into `auto_start`, hold an OAuth token, and are not under a switch-grade
kick block. Queue order follows the fallback chain, then display order.

Provenance, stated plainly: the implementation was authored across prior agent
sessions under the same operator, then revised against the maintainer's review.
The operator directed each stage and validated the feature live on their own
accounts. The checks below cover the rebased, review-complete diff.

User-visible changes, per the contributing guide:

- The Config tab adds `auto-start queue`, default off.
- Queue members show `[ queue n/N ]` on the Fallback card only.
- A landed queued open logs `5h auto-start window opened (queue n/N, ...)`.
  The line and anchor are gated to the lapsed leg, so a successful health
  re-test during a live window cannot claim an opening that did not happen.
- `status.json` adds `profiles[].auto_start_queue` with `position` and
  `next_open_at`; the object is null when the queue is off or the profile holds
  no slot, and `next_open_at` is null when the queue is due now.

Evidence from a live run (operator's account A and account B, N=2, about 15h):
seven consecutive openings alternated at about 2h30m — 17:10 → 19:40 → 22:13
→ 00:45 → 03:15 → 05:45 → 08:15 UTC. The wider 25h / 2,904-sample history
dataset showed no frozen-future idle reset at any offset, backing the
two-sample sliding classifier. The accepted failure trade is narrow: if a kick
lands, the follow-up `/usage` 429s, and the process crashes before any fresh
body, the restart forgets that opening and may spend one redundant ~1-token
kick; the next fresh pair self-heals the spacing. If the API later contradicts
the observed series behaviour, the fallback is a per-profile stamp written
through the existing `write_profile_cache` path.

## agent

- Claude Code on Nimbalyst (claude-fable-5), operated by @tmanaud.
- Ran `git rebase origin/mommy` (clean, no conflicts), `cargo fmt --all`,
  `cargo fmt --all -- --check`,
  `cargo clippy --all-targets --all-features -- -D warnings`,
  `cargo test --bin clauth auto_start_queue -- --nocapture` (25 passed), and
  `cargo test --all-targets --all-features --quiet` (main binary: 2,561 passed,
  0 failed, 3 ignored; other target suites: 2 passed and 8 passed). Release
  checks and the 3-OS matrix are left to CI.
- Operator verified the queue live on their own two accounts; account names and
  emails are intentionally omitted here.


uwuclxdy:
oooo this is so cool, thanks! will review in a bit

uwuclxdy (changes requested):
Thanks for the writeup and the live run, that made this much easier to review. The concurrency skeleton is right: lock ranks are correct at every acquisition site, electing in `tick` before the fan-out is the correct place to serialize it, and the leaf logic in `warmup.rs` is well tested. 21 of 22 mutations I planted against it died on the test you'd have named.

Trial-merged onto current `mommy` (you're 10 commits behind), conflict-free, and the gate is green there.

Requesting changes. A fair chunk of it is me cutting scope rather than you having done something wrong.

## Decisions that remove work

**1. `warmup_stagger` defaults OFF.** The doc comment argues default-on is safe because N=1 is bit-for-bit unchanged. That isn't true, see blocker 1. With it off the argument stops being load-bearing and nobody's behaviour changes on upgrade.

**2. The ring renders on the Fallback card only.** Drop the Overview `◎n/N` chip and the Setup `auto-start` hint arm. Two reasons for the chip: `◎` (U+25CE) is East-Asian-Ambiguous. The house rule for row glyphs is 1 cell everywhere (`⊘` got picked over `⚠` for exactly that). Second, the chip is the innermost trailer, so a ring member's blocked-reason marker sits 4 to 6 cells right of a non-member's, which breaks the shared trailer column the two-pass layout in `fallback_flow_lines` exists to build. `ChainRow::into_line` goes back to its previous form with it. I derived that your rewrite is byte-identical to the old one at every width for a row with no ring, so it wasn't a regression, it's just no longer needed.

The Setup arm sits above the plain on-state arm, so a ring member's `auto-start` row never renders `starts a throwaway session when idle so the 5h window counts`. The row loses its own help to a chain-wide status readout.

**3. No new file. Derive the anchor from `usage_history.jsonl`.** This is the big one. The module doc argues a persisted anchor is necessary because `/usage` can't distinguish an idle 5h window from a just-opened one. That's true of a single snapshot and not true of the series, which we already write per poll and keep for 2 days. An idle window's `resets_at` slides forward with the clock; a real one's is a fixed boundary. Two consecutive samples separate them. `profile::load_usage_history` is already the reader for it, with four existing callers.

That deletes `warmup_ring.json`, the `RingStamp` lock rank, `cold_fallback_open` and `implied_open_is_suspect`. It removes three findings I'm not going to bother listing, since they all live in code that goes away.

One caveat so you can push back on it: our own measurement pins the sliding behaviour to readings at exactly `now + 5h`. It says nothing about idle readings in the 3h to 4h band. If those turn out not to slide, we fall back to a per-profile stamp written through `profile_cache::write_profile_cache`, which `store_ring_stamp` currently re-implements line for line anyway (path, `serde_json::to_string`, `atomic_write_600`).

**4. One word for the concept.** Right now there are four: `stagger`, `ring`, `warm-up`, and `◎`. `warm-up`/`warmup` all become `auto-start`, which is the existing name for this. `ring` and `stagger` both become `queue`.

| yours | ours |
|---|---|
| `warmup_stagger` | `auto_start_queue` |
| `stagger warmup` (Config row) | `auto-start queue` |
| `warm-up` (Fallback card key) | `auto-start` |
| `[ ring 1/2 ]` | `[ queue 1/2 ]` |
| `5h warm-up window opened (ring 1/2, ...)` | `5h auto-start window opened (queue 1/2, ...)` |
| `"warmup": {"ring_position", ...}` | `"auto_start_queue": {"position", "next_open_at"}` |
| `src/usage/warmup.rs` | `src/usage/auto_start_queue.rs` |
| `WarmupRing` / `RingSlot` / `RingView` | `AutoStartQueue` / `QueueSlot` / `QueueView` |
| `ring_members` | `auto_start_queue_members` |
| `stagger_gap_secs` / `stagger_due` | `queue_gap_secs` / `queue_due` |
| `may_warmup` | `may_open_window` |

Root every new identifier at `auto_start_queue_` rather than a bare `queue_`. `feature_coverage.rs` matches with `name.contains(p)`. A bare `queue_` already hits `enqueue_refetch` and `requeue_quiet`, where `auto_start_queue_` matches nothing today. Same reason the current `ring_` entry needs to go: it matches 55 function names including `acquiring_an_outer_rank_while_holding_an_inner_one_panics`, so deleting every one of your tests would still leave that row green. `election_` does the same via `selection_caret_is_bold_like_every_other_card`.

`may_warmup` is the one deliberate exception to the single word. `collect_tokens` sets it true for paths that have no queue at all, so a name asserting election would be wrong in its default state.

## Blockers

**1. `note_ring_open` fires on a leg that opens nothing.** `scheduler.rs:1521-1529`. `kicked.opened` means a 2xx from `/v1/messages`, not that a window opened. On the live-window re-test leg `mark_window_open` correctly no-ops (its own doc: "No-op while the stored window is still live") yet the anchor moves anyway, re-phasing the whole ring for an event that didn't happen. The comment at the call site says "a window that opened is one the ring must space against". The comment 30 lines above it says the re-test "opens nothing". Both can't be right. Fix is to gate it on the `window_lapsed` value you already compute at `:1500` for `ring_open`.

This is why N=1 isn't bit-for-bit: an account with a standing kick block whose re-test lands mid-window pushes its own next warm-up out by as far into the window as the re-test landed.

**2. A non-member gets `may_warmup = false` forever.** `scheduler.rs:1393` and `:1409`. Both loops write over every entry in `due`, but `candidates` only iterates `ring`, and `ring_members` excludes switch-grade-kick-blocked and `auth_broken` profiles. So one of those profiles is denied the lapsed leg on every tick and never kicks again there. Since only a landed kick clears a block, it's stuck until the limiter's advertised ceiling passes and it falls out of `kick_rejected_names`. Before this PR it retried on the `kick_retry_due` ladder. Fix is to skip non-members in both loops rather than stamping them.

**3. The production call site isn't pinned.** `scheduler.rs:2741`. I wrapped the `elect_ring_warmup(...)` call in `if false` and all 2564 tests passed, plus clippy. Both `warmup_election_*` tests call `super::elect_ring_warmup` directly, so they pin the function and not the wiring. There are already four `super::tick(&state)` tests at `tests/inline/scheduler.rs:3938,4022,4137,4150` running under `EndpointSandbox`; one of them growing a ring would close this.

**4. The persistence and election-health half has no test references at all.** `note_ring_open`, `note_ring_kick_failed`, `ring_failures`, `store_ring_stamp`, `seed_ring_anchor`, `ring_anchor`, `next_ring_open_secs`, none of them appear anywhere under `tests/`. Deleting the calls is caught by clippy's dead-code lint only. Corrupting them isn't caught at all: I changed `note_ring_open(warmup, &entry.name, now_secs)` to pass `0`, which pins the anchor to the epoch so the ring never gates again, and it passed both the suite and clippy. Same for keying the failure streak on the wrong profile name. Most of this shrinks once decision 3 lands, though `run_fetch` needs a test entry point either way.

## Smaller, still worth fixing

- `should_open_window`'s new `ring_due` argument is inert to the suite. All 8 rows of `kick_suppressed_during_rate_limit_streak` pass `true` and no row was added. `!should_open_window(0, true, true, false, false)` is the missing one.
- `MAX_ELECTION_FAILURES` is unpinned in both directions. The fixtures are written `MAX_ELECTION_FAILURES`, `- 1` and `+ 3`, so they move with the constant and 2 to 99 survives. Write `2`, `1` and `5` as literals.
- `note_ring_kick_failed` only runs inside the `auto_start_should_kick` branch, so a member elected and then refused before the kick fires (a 429 streak, or a standing non-switch-grade block) records no failure and holds the slot on every tick. That is the head-of-line case `MAX_ELECTION_FAILURES` exists to prevent.
- `a_failure_streak_decays_so_a_skip_is_never_permanent`'s second half is captioned "once forgiven it takes its ring slot back" but calls `elect_warmup` with a hardcoded `0` and never touches `decayed`, `ring_failures` or `note_ring_kick_failed`. It's byte-equivalent to case 1 of `election_picks_one_lapsed_member_in_ring_order`.
- The test leg commented "a wrongly-sized ring would elect and this catches it" doesn't catch it. Sizing the gap off `due.len()` instead of `ring.len()` survives, because `due` is a subset of `ring` so the mutant's gap is always larger, and the fixture only asserts the hold direction. Flipping that leg's anchor to `Some(now - gap - 60)` and asserting the elect direction fixes it.
- `RingView` has no test references either. Making `slot()` always return `None` removes every chip on every surface and nothing reds.
- The `stagger warmup` Config row has no render or toggle test. Neither of its two hint strings is pinned. `RefreshSpentAccounts` and `PreemptiveRotation` both have one in the same files.
- `status.json`'s new object is pinned by presence in a key list only. Publishing `null` for every position survives.
- `ring_chip_renders_the_slot_and_the_rings_next_open` uses `row.contains(...)`; the house rule for a new surface pin is exact-match on the full row vec. Its own sibling `auto_start_hint_names_the_ring_slot_for_a_member` uses `assert_eq!` and says why.
- `usage/mod.rs:30` says the gap arithmetic stays module-private. `stagger_gap_secs` is re-exported four lines below.
- `warmup.rs:239`'s comment says `usize::MAX - 1` "leaves the name tiebreak room to decide without overflowing the key", but no arithmetic is ever done on that rank, so it and `usize::MAX` are identical on every input.

## Wiki

Don't spend time on this one, we'll rewrite both sections at merge since the rename touches every paragraph. Three claims to know about, because two of them are about the code and not the prose:

- "a quarantined, kick-rejected or spent one does not hold a slot open". `ring_members` has no spend, exhaustion or cancellation filter. `fallback::is_canceled` takes a plain `&Profile` and is the one that's missing.
- "spacing of at least `5h / N`". `stagger_gap_secs` subtracts the tolerance, so the enforced floor is 8910s at N=2, which is below `5h / N`, not at it. `profile.rs:539` gets the direction right.
- "a deleted `warmup_ring.json` re-spaces the ring over the next cycle". `ring_anchor` returns the in-memory value first, so a running daemon or TUI never notices the deletion. True only after a restart. Moot under decision 3.

`Configuration.md:111` still says the ping fires "on each refresh tick", which this PR makes false three paragraphs above its own correction.

Two committed pages own contracts this extends and weren't touched: `wiki/Daemon.md` is the `status.json` read contract, so it needs a row for the new object covering all three of its null cases. `wiki/Interface-And-Keys.md` lists every Config row down to `refresh spent`.

## README

Keep it on the existing bullet, no em-dash, and drop the trailing "so a freshly reset account is always close at hand":

```md
- 🤖 **Auto-switch** down a fallback chain the moment an account hits its limit, with weekly-window and spend-ceiling gates so a long run never stalls and never surprises you with a bill. Opted-in accounts queue their auto-start, opening 5h windows `5h / accounts` apart instead of all at once
```

tmanaud:
Thanks for the detailed review. I adopted all four decisions and addressed the
four blockers plus the smaller test and documentation items.

## Decisions

1. `auto_start_queue` now defaults off. Existing users keep the pre-PR
   auto-start behaviour until they opt in, and false is omitted from the saved
   global config.

2. The queue renders on the Fallback card only. I removed the Overview marker
   and restored the Setup `auto-start` row's original per-profile help. The
   remaining chip is `[ queue n/N ]`.

3. I removed `warmup_ring.json` and all dedicated stamp I/O. The queue anchor
   now comes from the existing per-profile `usage_history.jsonl` series: a
   fixed reset boundary confirmed by two samples is an opening, while a reset
   that advances with the sample clock is the idle shape. The scheduler seeds
   that result into memory before spawning and advances it on a landed lapsed
   kick; single-shot status derives it from history, while the daemon publishes
   the scheduler's own anchor.

   The empirical support is the 25h / 2,904-sample dataset from account A and
   account B: it showed no frozen-future idle readings at any offset, which
   backs the two-sample slide test. I accepted the remaining trade explicitly:
   if a kick lands, its follow-up `/usage` 429s, and the process crashes before
   any fresh body, the restart forgets the opening and may issue one redundant
   ~1-token kick; the next fresh pair self-heals. If the API later contradicts
   this behaviour, the fallback remains a per-profile stamp written through
   `write_profile_cache` rather than a separate write path.

4. The concept is now `auto-start queue` throughout: configuration,
   identifiers, module/test filenames, lock rank, log text, TUI chip, and the
   `status.json` object. Feature tests all use the `auto_start_queue_` prefix.

## Blockers

1. `run_fetch` captures whether the stored 5h window was lapsed before the
   kick. `note_queue_open` and the queue-open log run only for that lapsed leg;
   a successful live-window health re-test leaves the anchor alone.

2. Both election-stamping loops now touch queue members only. A non-member
   keeps `collect_tokens`' permissive `may_open_window` value and continues on
   its existing `kick_retry_due` ladder.

3. `auto_start_queue_election_is_wired_into_tick` runs two lapsed opted-in
   accounts through `super::tick` and the loopback endpoints. It asserts one
   messages kick, a populated shared anchor, and a live window for the elected
   first member, so removing the production election call produces two kicks.

4. The persistence and health paths now have direct pins. A `run_fetch` test
   proves a live-window re-test neither moves the anchor nor emits the queue
   opening line, then proves the lapsed leg stores a wall-clock anchor. A 403
   kick test proves the failure streak is keyed to the elected member rather
   than its sibling. A file-level JSONL test proves startup and cold fallback
   recover a confirmed boundary, reject a sliding idle series, and feed
   `next_queue_open_secs`.

## Smaller items

- Added the missing queue-gate false row to the `should_open_window` truth
  table.
- Pinned the election failure threshold with literal 2 / 1 / 5 fixtures.
- An elected lapsed member refused before the HTTP kick now records a failure,
  covering both a 429 streak and a standing retry clock.
- Reworked the failure-decay test through `note_queue_kick_failed` and
  `queue_failures` instead of a hardcoded count.
- Flipped the wrong-sized-membership fixture to the elect direction.
- Added a direct `QueueView` resolver test.
- Added an exact Config-row test covering both hint strings.
- Pinned `status.json` queue positions and all three null shapes.
- Changed both queue-chip assertions to exact full-row matches.
- Fixed the `usage/mod.rs` export comment and the `usize::MAX` tie-break
  comment.
- Added canceled subscriptions to the shared queue-membership exclusion.

## Docs and checks

I replaced the README feature sentence verbatim, documented
`profiles[].auto_start_queue` and its null cases in `wiki/Daemon.md`, and added
the `auto-start queue` Config row to `wiki/Interface-And-Keys.md`. I left
`wiki/Configuration.md` and `wiki/Auto-Switch.md` alone for the merge-time
rewrite you mentioned.

Commands run:

- `cargo fmt --all` — passed.
- `cargo fmt --all -- --check` — passed.
- `cargo clippy --all-targets --all-features -- -D warnings` — passed with no
  warnings.
- `cargo test --bin clauth auto_start_queue -- --nocapture` — 25 passed, 0
  failed.
- `cargo test --all-targets --all-features --quiet` — main binary: 2,561
  passed, 0 failed, 3 ignored; the other target suites passed 2/2 and 8/8.

After that run, a final review pass caught two new test fixtures whose
raw-string line continuations made the served `/usage` body invalid JSON — the
legs still passed, but through the parse-error path instead of a fresh body.
Fixed via `concat!`, then re-ran: `cargo fmt --all -- --check` clean,
`cargo clippy --all-targets --all-features -- -D warnings` clean, and the
focused queue suite 25/25.

Everything is amended into the single commit (63d6c56) and force-pushed with
lease. The commit message's anchor paragraph now describes the history-derived
anchor, and its surfaces paragraph the Fallback-card-only rendering.

## agent

- Claude Code on Nimbalyst (claude-fable-5), operated by @tmanaud; part of
  this round was implemented by a second agent session under the same
  operator, and every change was reviewed and re-verified by this one.
- ran: the commands above, plus the amend and `git push --force-with-lease`.
  operator verified: approved each decision (including the history-derived
  anchor) and this reply before posting.
- unsure: none beyond the trade named under decision 3.

uwuclxdy (changes requested):
All four blockers from last round landed: the anchor and the log gate on the pre-kick `lapsed` capture, both stamping loops touch members only, the election wiring is pinned (wrapping the `tick` call in `if false` reds it), and the persistence half has direct pins. Trial-merged onto current `mommy` and the full gate is green there.

One blocker on this round.

**The two-sample classifier self-confirms the idle shape on real files.**

`append_usage_sample` writes a bridge line whenever a sample changes: the previous sample re-stamped one ms earlier (`src/profile.rs:1291-1330`). An idle profile's `resets_at` slides with every poll, so every idle poll writes a bridge.

`confirmed_boundary` accepts any pair with `|dr| <= 61` that is not sliding. The pair `(fresh@T-90, bridge@T-1)` always has `dr = 0`, because the bridge duplicates the earlier sample byte for byte. The slide test cannot catch it: with `dr = 0` it would need `dt <= 15`, but it only engages at `dt >= 30`. So that pair confirms at every cadence, and `series_open` returns a recent sample time for any idle history with at least two polls. `history_anchor` takes the max over members, so one idle member seeds the anchor to ~now and `queue_due` then holds for up to one full gap after every spawn. At N=1 that is up to ~5h of suppressed auto-start on a lapsed window, silently: the chip and `status.json` show a plausible `next in`.

Both classifier tests hand-write series without bridge lines, a shape the writer never produces, so they pin a fixture rather than the file. I confirmed it with a probe replaying the writer's shape: `series_open` on `(t, t+5h)`, `(t+89, t+5h)`, `(t+90, t+90+5h)`, `(t+179, t+90+5h)`, `(t+180, t+180+5h)` returns `Some(t+90)`, where the module doc says an idle tail yields no anchor at all. The accepted crash trade reads differently on real files too: the restart does not forget the opening and elect immediately, the pre-crash idle bridges re-anchor ~90s before the forgotten kick, so the redundant kick usually does not happen. The scenario that fails is the restart with an idle history.

Fix directions:

- mark the bridge line in the writer and have the series reader drop it, or
- treat a pair whose two usage payloads are byte-identical as non-evidence. The bridge is a byte-identical re-stamp; a real kicked pair differs in at least one field.
- Requiring `dt >= SLIDE_BASELINE_SECS` for confirmation would work, but it collides with the pinned same-instant synthetic/floored confirmation (`auto_start_queue_series_open_confirms_kicked_windows_and_distrusts_a_single_reading`, where `series[..3]` expects `Some(kick)`). That path is open only if the merge write starts stamping distinct timestamps.

Seed the classifier tests through `append_usage_sample` itself, so the pin covers the writer's real output. The module doc's trade paragraph then needs its correction folded in.

Three more asks, all small.

1. Drop the tooltip under the queue row (`src/tui/render/chain.rs:581-583`). The pill and the Config hint already say it. The row sits above the fold on every member's card.
2. Right-align the countdown. `queue_pill_spans` appends `next in 2h 28m` inline after the pill, and at narrow card widths the trailing ETA is what clips: `next in 2h 26m` read `next in 2` at 60 columns on the earlier head. Right-aligned, the clip lands on the leading words and the digits survive at every operable width. The chip pin should render a narrow width too, not only 60.
3. Dim the Config row and no-op its toggle when no account has `auto_start` on. `GlobalConfigRow::AutoStartQueue` renders a plain `toggle_row` (`src/tui/render/global_config.rs:426-428`) and the handler flips unconditionally (`src/tui/app.rs:4478`). The file's own contract dims rows another setting makes inert: `clock`, `extra usage spent`, the burn rows. A live row with nothing behind it reads as a working control.

Copy, one line: replace the off-state hint `let every auto-start window reopen the moment it lapses` with `auto-start usage windows as soon as possible`, and write `5h / accounts` with spaces in the on-state hint to match the README and wiki.

tmanaud:
**Blocker**
Your second direction, strengthened twice: from pairwise non-evidence to
dropping the bridge outright, then from whole-payload identity to the
classifier's own projection. member_series discards any entry whose
five_hour serialization is byte-identical to its predecessor's.

Identity alone — my first cut — has a counterexample ten lines above the append
site: the /profile-despite-429 overlay advances plan in place with no
history line, so the next bridge re-stamps the SAME window under a changed
payload and confirms at dr = 0. Judged on the 5h window alone every re-stamp
drops, while the one bridge carrying new window information — the synthetic
mark_window_open stamp — differs in five_hour and survives, so the
same-instant synthetic/floored confirmation is untouched.

Dropping rather than skipping matters because of the OTHER pair:
(bridge@T-1ms, fresh@T) is byte-different with dt ≈ 0 and dr = +cadence,
so pairwise non-evidence leaves it confirming at any interval ≤ 61s, and
MIN_REFRESH_INTERVAL_MS is 10s. With the bridge gone it collapses to
(fresh, fresh), which slides.

That range exposed an adjacent pre-existing hole: below SLIDE_BASELINE_SECS
the slide test cannot engage, so a bare idle pair at a 10–29s cadence confirmed
too (dr = +dt ≤ 61). Below the baseline confirmed_boundary now confirms only
inside the one-way kick band [-BOUNDARY_AGREE_SECS, +1], where an idle pair
has dr = +dt ≥ +10; past the baseline is unchanged.

The pins run through append_usage_sample itself via a time-injected
append_usage_sample_at seam, so they cover what the writer actually produces —
one asserts the bridge lines reached disk (3 samples → 5 lines) before asserting
the idle series yields no anchor. The tuple tests stay hand-written on purpose:
they pin the pair arithmetic on post-member_series input, where bridges are
gone by construction.

**The three asks**
Tooltip dropped; the trailing blank stays.
The countdown right-aligns at the card's inner width and sheds its leading
   words before any digit (next in 2h 28m → 2h 28m), exact-matched at 60
   columns and at 34. queue_pill_spans is gone — the Fallback card was its
   only consumer.
The Config row dims and its key no-ops while no account has auto_start on
   (dimmed_toggle_row, the extra usage spent contract). Pinned all-faint
   while inert, live knob once an account opts in, and space no-opping then
   toggling-and-persisting.

Both hint strings replaced as given. Rebased onto current mommy through the
ProfileName, MemberCard, and build_profile_entries refactors; the
status.json object is now a typed QueueEntry on ProfileEntry
(#[serde(default)]).

Three more, found reviewing my own answer
The anchor only ever saw our own kicks. queue_anchor returned the
in-memory value the moment it was Some, and only a landed lapsed kick writes
it — so after the startup seed the derivation never ran again, and a window
opened by anything else was invisible to the gate for the life of the process.
The member behind it could then be elected seconds later, re-collapsing the two
windows the queue had just spaced. The rule is "no other member OPENED one",
not "no other member was kicked by us", so the gate composes both by max.

There are two unqueued openers, not one. Besides a live Claude Code session,
clauth use primes its target through prime_window — the second caller of
auto_start_kick — consulting neither the toggle nor the anchor. That override
is deliberate: a use that silently declined to prime is the worse lie, and the
CLI is a one-shot process with no share of the queue's memory. It costs one
interval of spacing, not the invariant, because the primed window lands in
history and the next tick re-anchors on it. Before this fix that was false — a
manual prime was invisible permanently. The module header now says all of this;
a comment of mine that claimed switching cannot open a window is reverted.

The replay had to stay off the common tick. Three bounds. A cached anchor
inside the gap returns as-is (max only moves an anchor forward, and forward is
deeper into the gap). The election builds candidates BEFORE reading the anchor,
so a tick with nothing lapsed answers without one — that covers the long middle
of every cycle. Neither covers an outage, where a member's kick and its /usage
refresh both keep failing, so it stays lapsed and the election keeps probing it
on purpose; the derivation is therefore memoized against a fingerprint of the
history files (mtime folded with length — a stat per member, not a parse). No
fresh body lands during an outage, so nothing is appended, so that is exactly
when the fingerprint holds still.

status.json published a queue the election was not running. The feed
passed an empty blocked set to the shared membership rule while the scheduler
passes kick_rejected_names and the TUI passes the switch_grade_kick_lifts
keys. My comment called the lag "one poll longer"; it is the limiter's
advertised ceiling, hours, and for that span a blocked account holds a position
and inflates N, publishing everyone else a next_open_at sized off a gap the
election never applies. Now symmetric with the anchor: a live daemon hands over
its in-memory blocks in LiveSignals, and status --json re-derives them from
the kick_block.json caches sync_kick_blocks_from_cache already reads.

Plus three stale comments: the Config row and the N=1 test still said the
feature defaults ON, and QueueView's doc still named the Setup and Overview
consumers round 1 removed.

**Checks**
fmt clean; clippy clean debug and --release; `cargo test --all-targets
--all-features` 2,758 passed / 0 failed / 3 ignored, plus 2/2, 1/1, 8/8, re-run
in full after every fix. Each was mutation-checked rather than eyeballed: with
the fix reverted under its own test the pin fails while the rest of the suite
stays green — which is why none of these showed up before. The memo's pin is
the invalidation one, since a stale memo would silently undo the fix it is
bolted onto.

The --release TEST leg is still not run: this machine's disk sits at 98% and a
full release codegen filled it last round. Release COMPILATION is covered above;
release execution rests on your trial-merge gate.

Amended into the single commit and force-pushed with lease.

**agent**
Claude Code on Nimbalyst, operated by @tmanaud. Rounds 1–2 answered with
  claude-fable-5; the fixes above written with claude-opus-5, then put through
  an adversarial read by gpt-5.6-sol (Codex). Every change re-verified by the
  operator.
ran: the checks above, the rebase, the amend, and `git push
  --force-with-lease`.
The cross-model pass earned its keep: it caught the clauth use claim I had
  just written into a comment, and the outage case my replay bound missed.
  Both are fixed above rather than disclosed.
unsure, two. The clauth use override is a judgement call, not a derivation;
  if you would rather prime_window consulted the gate, that is small and your
  call. And history_anchor is not clamped to <= now — a valid fixed 5h
  boundary implies boundary - 5h is past, so this is not a nominal defect, but
  a clock rollback or malformed upstream stamp could hold the queue until
  future + gap. Say the word and I will harden it.
The decision-3 trade stands as named, honest in both directions: an idle
  bridge can neither stand in for a forgotten open nor hold the queue shut
  after a restart.

uwuclxdy:
Landed as `3fd89866` on `mommy` (squash, with the `LiveSignals` test drift against current mommy resolved in the same commit). Thanks for the three rounds and the live run.

The rest is on our side: the two-sample classifier still self-confirms the idle shape through minute quantization at sub-2-minute refresh cadences (the 90s default included), and it gets reworked to span-based confirmation in a followup commit, together with the `wiki/Daemon.md` past-stamp clause and the `Configuration.md` / `Auto-Switch.md` rewrite. Closing this so the branch history doesn't fork.

PR #69, opened 2026-08-31
feat(codex): the codex harness, implemented as the six-part series

xingfanxia:
Implements `docs/codex-plan.md` end to end — the six-part series, cut from `mommy` and rebased onto `eb6cbb8`. Supersedes the implementation half of #51, which stays the design thread.

28 commits, every one of them building on its own (`git rebase --exec 'cargo check'`). 2815 debug / 2812 release passed, 0 failed, macOS arm64. Only clippy red in the tree is `src/tui/render/footer.rs:317`, which is byte-identical to `mommy` — my clippy is 1.96 and stricter than yours, so I left it alone rather than put an unrelated fix in this series.

## The six phases

| # | commits | what |
|---|---|---|
| 1 | 1–7 | harness axis: the `Harness` type, `codex-profiles.toml`, cross-file name uniqueness, `reload_fingerprint`, the live-row tag, `which`'s codex arm, switch/delete off the bare name — plus folded fixes 2/4/5 |
| 2 | 8 | the two engine seams, claude behind them, behavior-preserving |
| 3 | 9–13, 16 | per-session `codex-home-<sid>`, the forced file store, the scrub list, teardown sync-back, `login --codex` as ADOPT, `gc_codex_homes` |
| 4 | 14–15, 17–19 | the standby refresh under `RotationGuard` — durable no-replay memo, the belt, the 2-strike kick, the 5-minute stand-down — and the PKCE browser login |
| 5 | 26 | the `wham/usage` leg, the per-harness chain, folded fix 1, and `kick_codex`'s 401 producer |
| 6 | 27–28 | additive published fields, the copy fixes, the TUI codex section + `c` filter |

Commits 20–25 are a late sweep of their own; see "What the sweep found" below.

## Decisions as implemented

Everything under "Pinned decisions" holds as written, with three things worth calling out because the code says more than the spec could:

**Decision 6's premise is wrong, and I could not fix it here.** `-c` is the `SessionFlags` layer at precedence **30**; `LegacyManagedConfigTomlFromFile` is **40** and the MDM layer **50**, both pushed after it, and the unix managed path is the machine-global `/etc/codex/managed_config.toml` whose resolver explicitly discards `CODEX_HOME`. So on an MDM-managed Mac a `cli_auth_credentials_store = "keyring"` there silently defeats the forced override. Detail and a detect-and-warn proposal in #51 (issuecomment-5471926707) — I left the code alone pending your call.

**A forced `-c` alone is not enough even without MDM.** `debug.config_lockfile.load_path` makes codex rebuild the whole config from `ConfigLayerStack::new(vec![lock_layer], ..)` and return, so the `-c` layer is not outranked but **erased**. Since the session's `config.toml` is a verbatim copy of the operator's, an operator who set that key unpinned both overrides at once. The copy is therefore **sanitized** — `sqlite_home`, `cli_auth_credentials_store`, `debug.config_lockfile` dropped — with the forced `-c` kept as a second line. An unparseable config still copies through verbatim.

**Decision 8's concurrency premise does not hold under `LinkMode::Fake`.** The two flavors collapse to different bare stems, each holding its own copy converged from the same store — two carriers of one single-use chain, with no reload-and-skip between them. The acquire refuses the second *flavor* of one profile there; same flavor stays free, because that is one home.

## What the sweep found

After phase 4 I generalized one finding — codex resolves its state DBs through `sqlite_home`, a config KEY as well as an env var, so every profile's goals/logs/memories/state could land in one directory with the session home's durable links never opened through — into an adversarial sweep of the whole codex surface. 18 findings survived refutation; commits 20–25 are the fixes. The ones that would have shipped as real bugs:

- **`CODEX_API_KEY` / `CODEX_ACCESS_TOKEN` outrank the linked `auth.json`.** codex's `load_auth` resolves those before the persistent store and says so in its own comment. We scrubbed `OPENAI_API_KEY` for exactly this reason and had missed its two siblings, so `clauth start <p>` could authenticate as a different account while clauth reported and burned the named one. The three endpoint overrides join them, since a codex refresh token is single-use.
- **The stale-home wipe demanded the wrong thing.** It tested the shared bare stem only, but every Fake name is sid-free, so the isolated home — which holds a physical `auth.json` copy — was deleted after a crash and the spent store token converged back out.
- **`gc_codex_homes` decided liveness and removed outside the state lock**, on a comment that was false: the acquire flocks the marker *after* building the home, so the whole build window read as dead.
- **The capture refusal prescribed chain destruction.** codex's `login` opens with `logout_with_revoke`, which loads the stored auth *through the adopted symlink* and POSTs its refresh token to the revoke endpoint before minting — so "run `codex login` to mint a fresh chain" killed the already-captured profile server-side. Copy fixed on both surfaces; the underlying shape is decision 8's and is raised in #51.
- **`write_profile_cache` gated on `is_configured`**, which reads `profiles.toml` alone by design, so every codex usage reading was dropped silently — the fetch succeeded and the by-name cache stayed empty forever.

## Deviations, each deliberate

- **Folded fix 1** is answered by filling the freshness pass for codex rather than changing `is_exhausted_from_usage`. A member with no entry reading as un-exhausted is the only safe reading; what was missing is that it is not *proof* of headroom, and the freshness preference is exactly the mechanism for that. Changing the shared predicate would have moved claude behavior, which decision 9 forbids.
- **The TUI codex section is read-only and not selectable.** The cursor and every action are bound to `config.profiles`; a filter that could hide the selected row is a footgun. `c` cycles both → claude → codex → both and the header chip names the filter.
- **`sessions/` is left as the spec's runtime table has it.** The sync-back is write-only — nothing restores it, and the durable thread DBs hold absolute rollout paths into deleted homes. Making it a symlink into the store would canonicalize to one stable path across sessions, the way the DBs already do, but that contradicts your table, so it is a question in #51 rather than a change here.

## Verification

Each phase: implementation → adversarial review fleet → targeted mutation sweep → both test legs. 38/38 mutations killed across the series, including the ones that matter most — the weekly/5h cutoff constant, the window-collision rule, which window a `limit_reached` attaches to, the absolute-vs-relative reset precedence, and the cache gate above.

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -891,49 +891,85 @@ pub(crate) fn delete_codex_profile(name: &str, force: bool) -> Result<()> {
 }
 
 /// `clauth login <name> --codex` — create (or re-authenticate) a codex
-/// profile by capturing the operator's own `codex login` from
-/// `~/.codex/auth.json` into `profiles/<name>/auth.json`. The bytes are
-/// copied VERBATIM — a whole-file snapshot preserves every key, known or not
-/// — written atomic and 0600 (this writer owns that mode; the perms sweep
-/// deliberately does not descend into codex territory).
+/// profile by ADOPTING the operator's own `codex login`: the chain moves
+/// VERBATIM into `profiles/<name>/auth.json` (atomic, 0600 — this writer owns
+/// that mode), and the operator's `auth.json` becomes a symlink to it. One
+/// physical file is the design's own safety mechanism (decision 8): the
+/// operator's bare `codex`, every clauth session, and clauth's rotation all
+/// hold the same chain. A snapshot-copy here would be the forbidden
+/// configuration decisions 7/8 exist to prevent — two carriers of a
+/// single-use rotating chain, where the first refresh on either side strands
+/// the other. Where the operator slot cannot be linked (a host without
+/// symlink privilege), the copy is taken anyway and that exact hazard is
+/// said out loud instead of implied away.
 ///
-/// Two refusals, each naming its fix:
-/// - keyring/auto store (`cli_auth_credentials_store` in the operator's
-///   config.toml): under those modes the file is absent or stale BY DESIGN,
-///   so a capture would snapshot nothing or yesterday's chain.
-/// - no `tokens` chain in the file: an API-key-only codex setup has no chain
-///   to manage — rotation, usage, and the session symlink all speak ChatGPT
-///   tokens.
+/// The operator home is the one the operator's codex actually uses: a set
+/// `CODEX_HOME` is honored — unless it names a home clauth built, which means
+/// this shell is INSIDE a clauth codex session and "the operator's login" is
+/// some profile's store; that refuses rather than snapshotting a sibling.
 ///
-/// What capture deliberately does NOT do: touch the operator's `~/.codex` in
-/// any way. The cost is stated out loud at the end — the operator's own
-/// bare `codex` now holds a diverging copy of a single-use rotating chain,
-/// and whichever side refreshes first strands the other — because silence
-/// there would read as "both keep working", which is exactly false.
+/// Refusals, each naming its fix:
+/// - any store mode other than the file default (`keyring`, `auto`,
+///   `ephemeral`, or something newer): the file is absent, stale, or
+///   nonexistent BY DESIGN under those, so a capture would snapshot nothing
+///   or yesterday's chain. Allow-list, not deny-list — an unknown future
+///   mode refuses instead of guessing.
+/// - no `tokens` chain in the file (an API-key-only setup): nothing there
+///   for rotation, usage, or the session symlink to manage.
+/// - a slot already adopted by ANOTHER profile: one chain, one profile.
+/// - a live session on the target profile: re-capture replaces the chain the
+///   running session holds.
 pub(crate) fn codex_login_capture(name: &str) -> Result<()> {
-    let operator = crate::profile::home_dir()?.join(".codex");
-    let store_mode = codex_operator_store_mode(&operator);
-    if matches!(store_mode.as_deref(), Some("keyring") | Some("auto")) {
-        bail!(
-            "the operator codex stores its login in the OS keyring \
-             (cli_auth_credentials_store = \"{}\" in ~/.codex/config.toml), so \
-             ~/.codex/auth.json is absent or stale by design — set it to \
-             \"file\", run `codex login`, then re-run this capture",
-            store_mode.unwrap_or_default()
-        );
+    let trimmed = validate_name_chars(name)?.to_string();
+    let operator = codex_operator_home()?;
+    match codex_operator_store_mode(&operator).as_deref() {
+        None | Some("file") => {}
+        Some(mode) => bail!(
+            "the operator codex does not keep its login in auth.json \
+             (cli_auth_credentials_store = \"{mode}\" in {}/config.toml), so there is \
+             nothing current to capture there — set it to \"file\", run `codex login`, \
+             then re-run this capture",
+            operator.display()
+        ),
     }
     let auth_path = operator.join("auth.json");
+
+    // A slot clauth already adopted: the chain belongs to exactly one profile.
+    if let Ok(target) = std::fs::read_link(&auth_path)
+        && let Some(holder) = clauth_auth_store_owner(&target)
+    {
+        if holder.eq_ignore_ascii_case(&trimmed) {
+            outln!(
+                "clauth: {} already follows codex profile '{holder}' — nothing to capture",
+                auth_path.display()
+            );
+            return Ok(());
+        }
+        bail!(
+            "{} is already captured as codex profile '{holder}' — one chain, one \
+             profile. Run `codex login` to mint a fresh chain, then capture that \
+             into '{trimmed}'",
+            auth_path.display()
+        );
+    }
+
     let raw = match std::fs::read(&auth_path) {
         Ok(raw) => raw,
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
             bail!(
-                "no codex login to capture — ~/.codex/auth.json does not exist; run `codex login` first"
+                "no codex login to capture — {} does not exist; run `codex login` first",
+                auth_path.display()
             )
         }
         Err(e) => return Err(e).with_context(|| format!("failed to read {}", auth_path.display())),
     };
-    let parsed: serde_json::Value = serde_json::from_slice(&raw)
-        .with_context(|| format!("failed to parse {}", auth_path.display()))?;
+    let parsed: serde_json::Value = serde_json::from_slice(&raw).with_context(|| {
+        format!(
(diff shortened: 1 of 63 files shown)


uwuclxdy (changes requested):
Reviewed at the head. The spec held, the isolation sweep found and fixed real defects, and the wire matches codex's own client. Requesting changes: the guard gap, the unquarantined death path, two owner rulings, and one copy ruling.

The head does not compile against current `mommy`: `ProfileEntry` gained `auto_start_queue` and `PlanInfo` gained `codex_plan` after your base, so the codex entry literal (`src/daemon/status_json.rs`) and two mommy-side test literals need the new fields. A trial merge carrying those three fills gates green on our side (692s, 2938/2941 passed).

## Required changes

1. **The codex delete takes no rotation guard** (`src/actions.rs:868`). A standby rotation racing `remove_dir_all` either resurrects an orphan `auth.json` holding the new chain, or loses the minted pair after the old token was spent: `refresh_token_reused`, browser re-login only. `delete_profile` takes `rotation_guard_for_mutation` for exactly this race, and its doc names it; `delete_codex_profile` must too.
2. **The login callback echoes raw browser text** (`src/codex_login.rs:206-208`). The `error` query param rides `bail!` onto the operator's terminal unvalidated. The claude callback parses it into a closed set and discards the bytes; do the same.
3. **Dead chains never quarantine.** `Reused`/`Dead` classify, then `standby_pass` returns `Failed` and nothing records the verdict; past the breaker the member stays viable forever (`broken`/`kick_rejected` hardcoded empty in `snapshot_codex_chain`), so auto-switch keeps picking it and `clauth start` lands on a chain that cannot authenticate. The accepted design's "a genuinely dead chain exits through Permanent → quarantine" needs its exit.
4. **The new global `c` key kills the Tokens binding** (`src/tui/app.rs:2866`). The arm returns on every tab before per-tab dispatch, so `toggle_count_cache` is unreachable and the footer, help modal and wiki still advertise `c count cache`. The house shape branches inside the global arm per tab.
5. **Deleting an adopted profile leaves the operator slot dangling.** `delete_codex_profile`'s premise "nothing global is installed" is false: capture installs the `~/.codex/auth.json` symlink. After the delete, the operator's bare codex points at nothing and the delete says nothing. Detach the slot before removing the dir, and say so.
6. **An unparseable `codex-profiles.toml` pauses claude settings sync** (`src/settings_sync.rs:182-191`). The fix class this lands in exists to stop a codex file stalling a claude subsystem; the shipped mechanism re-opens it one level up. On an unreadable codex roster, treat the codex arm as absent instead of failing closed: a pure-codex dir carries no `[env]`, and dual claims resolve claude-first.
7. **Owner ruling: managed config can defeat the forced `-c`.** Verified against `rust-v0.145.0`: `SessionFlags` is precedence 30, `LegacyManagedConfigTomlFromFile` 40, MDM 50; `debug.config_lockfile.load_path` rebuilds the stack from one layer; the unix managed path is `/etc/codex/managed_config.toml`. A managed `cli_auth_credentials_store = "keyring"` silently unbinds the linked `auth.json` and the chain dies by replay. Your detect-and-warn proposal is wanted in this series: read the managed path at spawn and refuse or warn loudly when it sets the keys we force or the lockfile path.
8. **Owner ruling: `sessions/` becomes a symlink into the store**, like the sqlite DBs. The write-only sync-back leaves prior rollouts invisible to later sessions and prunable by codex's own listing. Concurrent codex instances already share one inode through the DB links, so the same pattern applies here. The runtime-table row changes with it.
9. **Owner ruling: the new codex-side copy says remove, not delete** (`clauth: removed codex profile …`, the live-session refusal). The `delete` verb and the shared confirm gate stay.
10. **The production wiring ships unpinned.** `fetch_codex_usage`, `standby_tick`, `apply_codex_switch` and the PKCE network legs (`exchange_code`, `exchange_api_key`, `bind_registered_port`) have no test through their entry points; each is deletable with the suite green. The doc claim that the network legs are covered "where the daemon standby tests exercise the shared agent" is false: those tests exercise `refresh_codex_chain_at`. Pin them, and give the wham leg an endpoint seam.

## Minors

- `rate_limit_reached_type` is never renamed to the slot its window landed in, and the `limit_reached` bump lands on the fuller window, which can be the wrong one (`src/usage/codex.rs:158-181`).
- Settled Q5's fallback half is absent: no id_token `chatgpt_plan_type` reader, so a never-polled account reads null tier.
- The memo write is fail-open: a failed `atomic_write_600` still sends the token with no durable memo (`src/codex_auth.rs:329-335`).
- The unstripped config copy lands 0644 in a tree the perms sweep exempts (`src/runtime.rs:4086-4088`).
- Fake-mode convergence breaks mtime ties toward the store, the permanent-death direction on coarse-mtime filesystems.
- The codex weekly soft line binds the claude `weekly_switch_threshold_pct`; no independent setting.
- `switch_profile` and the `profiles` filter still say "profile not found" for codex names; the spec's copy fix named both.
- The header chip count is always the claude count; codex-only view reads "3 accounts · codex only" over codex rows.
- In codex-only view the selected claude row is hidden and enter/arrows/`a` stay armed, so reorder acts blind; the body's own footgun rationale is what this violates.
- Codex rows past the viewport clip with no scrollbar (`total` counts claude rows only).
- Claude-only verbs (`disable`/`enable`/`rolling-token`/`static-token`) list codex names as available.
- `build_status` loads the codex state twice; `profiles[].active` can disagree with `active_codex_profile` in one feed.
- The client id is pinned by comparing the constant to itself; a wrong id ships green. Same family: the weekly-cutoff pin is green for any value in (5h, 7d]; the stand-down window is pinned only ≥120s.
- Please record in the body's deviations section the four mechanisms that deviate from the spec's letter: the sequential codex legs (not a third `thread::scope` arm), the dir-walk settings fix (not roster-driven), the absence of a per-harness pending-switch queue, and the kick bypassing the memo as well as the age gate.

The plan's old-daemon warning line is dropped by owner ruling; `clauth_version` stays additive. The copy drift in the new strings (em-dashes, caps emphasis, literal backticks in TUI prose, the `harness` label) and the wiki updates are handled maintainer-side at merge.

uwuclxdy:
Thanks for the series! Landed the review's asks on the branch (`88cb910d`, `3b10a4e7`, `dc169889`), the spec reconcile (`4b18b0a5`) and the wiki page plus README/SECURITY rows (`7ee6fc2e`), on top of a merge from `mommy`.

Two calls worth knowing: an unrecognized 4xx from the token endpoint stays at the memo/kick/breaker instead of quarantining (a WAF hiccup would otherwise lock every codex profile out with a fix line that cannot help), and the quarantine's fix line is `clauth login <name> --codex --browser`, since the bare capture is a no-op for an adopted slot. The two shipped strings that said `clauth switch <name>` lost the `switch` token: the switch is the bare `clauth <name>`. The rest of the copy pass lands separately.

`docs/codex-plan.md` carries the deviations from the spec's letter as shipped, incl. the capture re-stamping `last_refresh` (the fake-link convergence decides by it first) and the lockfile trigger being `debug.config_lockfile.load_path`.

PR #70, opened 2026-09-04
feat(providers): typed MiniMax provider, and let provider windows drive the chain

0xNayel:
Two changes. The first is a plain provider addition; the second is a behaviour change I'd rather you ruled on than merged silently, so it's easy to drop if you disagree — the provider stands alone without it.

## 1. Typed MiniMax provider

MiniMax matched no entry in the provider table, so a profile on `https://api.minimax.io/anthropic` fell through to `generic`: no typed fetch, no panel, `provider = "generic"` in the status feed.

`providers::minimax` fetches `GET /v1/token_plan/remains`. The response carries one entry per plan bucket, with the interval and weekly windows as **remaining** percentages, so the fetch inverts them into the utilization every other provider's bar already carries.

Judgement calls worth flagging:

- **Bars follow the `general` bucket**, the one Claude Code bills against. `video` shares the account but not that window, so it rides as a row only — bars sourced from it would report headroom the completions don't have. A lone bucket under another name drives the bars (an account entitled to one product can't be ambiguous); several buckets with no `general` yield no bars rather than a guess.
- **International host only.** MiniMax also fronts a China-region endpoint, but `Provider::fetch` resolves its origin from a constant rather than from the profile's own `base_url`, so claiming a second origin here would send one region's api key to the other region's host. A CN account keeps falling through to the generic scanner until `ThirdPartyTarget::Known` carries the base URL. Happy to do that plumbing in a follow-up if you want the region covered.
- **No plan label.** The response names no tier. The only "token plan" in it is the endpoint path, and a label built from a path puts a word in the vendor's mouth.
- `base_resp.status_code` is checked separately from the HTTP status: a rejected key answers `200` with a non-zero code, so the transport status alone would publish an empty plan as a healthy one.

Console URL is `platform.minimax.io/user-center/basic-information`, per their own docs. Preset added alongside.

## 2. Provider windows now reach the scheduling layer

This one is not MiniMax-specific and is the reason the PR is worth more than one file.

`ThirdPartyStats::to_usage_info` maps the two labels the chain judges (`5h`, `7d`) onto `UsageInfo`, and the third-party fetch mirrors that into the usage store the auto-switch walk reads.

Before this, `next_auto_switch_target`, `clauth list`'s used columns and `published_windows` all read one map that **only the OAuth leg ever wrote**. So a `fallback_threshold` set on an api-key member could never fire, and a Z.ai account in a chain rendered perfect 5h/7d bars on the Usage tab while the walk treated it as windowless. Same for Alibaba.

So this changes existing behaviour: a chain already holding a Z.ai or Alibaba member starts switching where it never did. That's the part I'd want your ruling on — it reads to me like the bars were always meant to mean this, but it is a live change to what an existing config does.

Three deliberate limits on the mapping:

- **Only `5h` and `7d`.** Any other label (z.ai's `30d`) is dropped rather than folded into `weekly_scoped`, which means per-**model** weekly windows and carries the `check_scoped` gate's semantics. A monthly account-wide ceiling landing there would block the member as though one model were capped.
- **`best_effort` stats never become windows.** The generic scanner's guess at an unknown endpoint is a figure nobody verified; parking an account out of rotation on one is worse than not switching. They still render as bars, which is a claim about the display only.
- **The mirror publishes to the usage store only.** `usage_cache.json` is still written by the OAuth leg alone, so a third-party account cannot resurrect a stale Anthropic window from an earlier OAuth life — the case the old `published_windows` early-return was guarding, which is now handled by reading each account from its own cache instead.

The cache-fallback leg cold-fills rather than overwrites, matching the rule the third-party store's own `or_insert` follows.

## Testing

18 new tests: the MiniMax wire shape parsed against a real captured response (bucket selection, the remaining→utilization inversion, reset instants, the danger row, absent/empty/ambiguous plans, the clamp), plus the derivation's label filter, its `best_effort` refusal and its clamp, and a `bootstrap_third_party` assertion that a seeded window reaches the map auto-switch reads.

`cargo clippy --all-targets` and `cargo fmt --check` are clean.

Note on the suite: on Windows a full parallel `cargo test` fails 2–4 timing-sensitive scheduler/oauth tests per run, a different set each time. I confirmed the same on unmodified `v0.15.1` before blaming my own change; every one of them passes in isolation. Not something this PR touches, but worth knowing if CI is ever red there.

## Verified against a live account

```
  PROFILE  PLAN     5H USED  7D USED  ENDPOINT
* pro      Pro          31%       7%  -
  minimax  MiniMax      73%      32%  https://api.minimax.io/anthropic
  glm-n    Z.ai          0%      49%  https://api.z.ai/api/anthropic
  glm-m    Z.ai          0%       1%  https://api.z.ai/api/anthropic
  max      Max 20x       0%     100%  -
```

The two Z.ai rows showed `-` in those columns before this change — that's part 2, not MiniMax. The MiniMax fetch was also driven against the real endpoint through `get_json` to confirm the whole chain: live response → bars → derived `UsageInfo` carrying both windows.

Wiki updated: `Configuration.md` (typed-panel table, console pages, preset table), `FAQ.md`, and `Auto-Switch.md` for the behaviour change in part 2.

https://claude.ai/code/session_01W8d4LcSFSoPuB5XSZE3TbY


src/providers/minimax.rs
```diff
@@ -0,0 +1,202 @@
...
+    super::url_matches_host(url, ORIGIN)
+}
+
+pub(super) fn fetch(api_key: &str) -> Result<ThirdPartyStats, ThirdPartyError> {
+    let text = super::get_json(&format!("{ORIGIN}{REMAINS_PATH}"), api_key)?;
+    let body: RemainsResponse = serde_json::from_str(&text).map_err(|_| ThirdPartyError::Parse)?;
+    // `base_resp.status_code` is MiniMax's own verdict and rides an HTTP 200:
+    // a rejected key answers 200 with a non-zero code, so the transport status
+    // alone would publish an empty plan as a healthy one.
+    if body.base_resp.status_code != 0 {
```
uwuclxdy: The in-band verdict can fail open. `base_resp` and `status_code` are both `#[serde(default)]`, so a 200 body without the envelope parses as `status_code: 0` = success and publishes a healthy empty plan: cache overwritten, `Fresh`-stamped, and with part 2 the walk windows get REMOVED. Every sibling fails closed on a missing verdict: z.ai's `success` has no serde default (missing field = parse error), Alibaba's defaults false and rejects. Make the verdict non-defaulted and reject a missing one. Separately: a rejected key's in-band `2049 invalid api key` maps to `ThirdPartyError::Status` -> `Failed`, so a dead MiniMax key is never suppressed, never records `third_party_auth.json`, and never renders `(key rejected)` - the wiki promises "A dead api key stops polling the same way, on any endpoint". Map auth-class codes to `AuthExpired`, the way Alibaba's HTTP-200 body verdict already does. What did a wrong key actually return in your live testing, HTTP status and body?

src/providers/minimax.rs
```diff
@@ -0,0 +1,202 @@
...
+use super::{StatRow, StatRowKind, ThirdPartyError, ThirdPartyStats, UsageBar};
+use crate::usage::{LABEL_5H, LABEL_7D, epoch_secs_to_iso};
+
+pub(super) const DISPLAY_NAME: &str = "MiniMax";
+
+pub(super) const ORIGIN: &str = "https://api.minimax.io";
+
+/// Where an operator mints the api key this provider authenticates with, as
+/// published by <https://platform.minimax.io/docs/guides/text-generation>.
+pub(super) const CONSOLE_URL: &str = "https://platform.minimax.io/user-center/basic-information";
```
uwuclxdy: This page is the pay-as-you-go key page. The provider reads the Subscription Key (`sk-cp-...`), minted at `platform.minimax.io/user-center/payment/token-plan` (quickstart-preparation.md splits the two, and text-generation.md's own `api_key` row links the token-plan page). The wiki row at Configuration.md:45 needs the same URL.

src/providers/minimax.rs
```diff
@@ -0,0 +1,202 @@
...
+    }];
+    for m in models {
+        let pct = |p: Option<i64>| match p {
+            Some(p) => format!("{p}%"),
+            None => "-".to_string(),
+        };
+        rows.push(StatRow {
+            label: m.model_name.clone(),
+            value: format!(
+                "{LABEL_5H} {}  ·  {LABEL_7D} {}",
```
uwuclxdy: Two things on this line. Every bucket prints the `5h` shorthand, but the `video` bucket's interval is 24h in both real captures (end - start = 86,400,000 ms), so the row claims a window length the bucket does not have. Derive the length from the timestamps, or drop the shorthand for non-`general` buckets. And the `  ·  ` double spacing: the house separator is single-spaced ` · ` everywhere else.

src/usage/scheduler.rs
```diff
@@ -2129,6 +2130,37 @@ fn try_seed_cache(
...
+/// from a window (a live percentage against a threshold), and the derivation
+/// declines anything it cannot vouch for, so nothing here branches on provider.
+///
+/// `None` REMOVES rather than leaves the previous entry. A provider that
+/// stopped publishing windows — a plan change, a response shape clauth no
+/// longer recognises, a best-effort scan that used to guess one — has no
+/// current reading, and a frozen entry would keep answering the walk with a
+/// figure nothing refreshes: exactly the stale-cache trap the OAuth leg's
+/// liveness checks exist to avoid.
+fn publish_third_party_windows(
```
uwuclxdy, line 2476: The `None` arm is the stale-cache guard half of this function's own doc, and no test drives it: the only added scheduler test asserts the bootstrap insert. Also unpinned: the fetch legs' two store writes, the `published_windows` third-party branch, the `load_profile` seed, and the tui re-derivation. At minimum pin the removal arm.

wiki/Configuration.md
```diff
@@ -85,12 +87,13 @@ A preset is a named `base_url` + `[models]` pair you can stamp onto any account
...
 
-`DeepSeek`, `Z.ai` and `OpenRouter` set the endpoint plus a base model, leaving the tier rows yours to pin afterwards. The four Alibaba ones fill every row instead, because those endpoints reject a Claude model id outright rather than serving something for it, so any alias left unpinned fails on use. All seven leave the api key alone; pick the region your plan was bought in, since a key issued for one is not accepted by the other. Once a preset is stamped on, `open provider console` in the same menu opens that endpoint's own key page ([above](Configuration#where-the-keys-come-from)).
+`DeepSeek`, `Z.ai`, `OpenRouter` and `MiniMax` set the endpoint plus a base model, leaving the tier rows yours to pin afterwards. The four Alibaba ones fill every row instead, because those endpoints reject a Claude model id outright rather than serving something for it, so any alias left unpinned fails on use. All eight leave the api key alone; pick the region your plan was bought in, since a key issued for one is not accepted by the other. Once a preset is stamped on, `open provider console` in the same menu opens that endpoint's own key page ([above](Configuration#where-the-keys-come-from)).
```
uwuclxdy, line 96: Eight now: the table below has 8 rows and this paragraph says "All eight", but the section above still says "Seven ship built in:".

wiki/Configuration.md
```diff
@@ -21,26 +21,28 @@ A mint is a narrower credential than a `/login` session: it carries `user:infere
...
+Five providers get typed usage panels:
 
 | Provider | Base URL | Shows |
 |----------|----------|-------|
 | DeepSeek | `https://api.deepseek.com` | balance rows per currency: api balance, granted, topped up |
 | Z.ai | `https://api.z.ai` | percentage bars per limit window (5h / 7d / 30d), per-tool rows, plan level, 7-day per-model token totals |
 | OpenRouter | `https://openrouter.ai` | wallet rows from the credits endpoint: api balance (remaining credits, red when overdrawn), used, purchased; then today / this week / this month usage, per-key cap rows when set, free-tier flag |
 | Alibaba Model Studio | the four Qwen preset endpoints below | a 7d bar carrying your tier's absolute allowance, a 5h bar when the API reports one, plan tier, subscription status and days left |
+| MiniMax | `https://api.minimax.io` | Token Plan bars for the 5h interval and the 7d window, plus a remaining row per plan bucket. The bars follow `general`, the bucket Claude Code bills against; `video` and any other bucket ride as rows only. The mainland-China endpoint is not covered — it is a separate account on a different host, so it falls to the best-effort scan below |
```
uwuclxdy: Not true for a lone bucket: `bar_bucket` falls back to a single bucket of any name, and `a_lone_bucket_drives_the_bars_even_when_it_is_not_named_general` pins it. Something like "with more than one bucket and no `general`, no bars are drawn".

src/tui/app.rs
```diff
@@ -2370,6 +2371,18 @@ impl App {
                 {
                     p.third_party_usage = s.get(p.name.as_str()).cloned();
                 }
+                // A third-party member's bars are its usage snapshot. The OAuth
+                // store above just wrote `None` over this account (it holds no
```
uwuclxdy: This comment describes a store state the mirror no longer produces: after the first fetch, `s.get(p.name)` holds the mirrored entry, so the `is_none` re-derivation is dead on the second apply and the seed is never "gone by here" for a cached account. The comment is the only place the hybrid precedence rule is documented, and it documents a mechanism that does not exist.

src/providers/mod.rs
```diff
@@ -161,7 +166,7 @@ impl Provider {
     /// has none: a windows-publishing provider HAS the limits even when one
     /// cached response carried no bars.
     pub(crate) fn publishes_windows(self) -> bool {
```
uwuclxdy, line 169: `publishes_windows`' doc still reads "Zai and Alibaba do; DeepSeek and OpenRouter publish a wallet", and `console_url`'s says "the other three have one page each". Both went stale with this PR. Same for `ProfileEntry.windows`' "The OAuth 5h/7d usage rows" and mcp/render.rs's "(z.ai, Alibaba)".

src/profile_json.rs
```diff
@@ -239,20 +239,28 @@ pub(crate) fn oauth_windows(usage: &UsageInfo) -> Vec<Window> {
...
+/// is what makes publishing the third-party figures safe rather than a second
+/// way to render that same leftover.
 pub(crate) fn published_windows(name: &ProfileName) -> Vec<Window> {
     if crate::profile::stored_usage_cache_is_third_party(name) {
-        return Vec::new();
+        return load_profile_cache::<ThirdPartyStats>(name, THIRD_PARTY_CACHE_FILE)
+            .as_ref()
+            .and_then(ThirdPartyStats::to_usage_info)
+            .as_ref()
+            .map(oauth_windows)
```
uwuclxdy: `oauth_windows` now serves third-party windows too; the name and the doc at its definition claim OAuth-only. Rename to `usage_windows` (3 sites: the def and these two callers).

src/providers/minimax.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Remaining percentage → utilization, clamped into 0..=100. `None` for an
+/// absent figure, so a bucket reporting nothing yields no bar rather than a
+/// full-headroom one.
+fn utilization(remaining_pct: Option<i64>) -> Option<f64> {
+    remaining_pct.map(|r| (100.0 - r as f64).clamp(0.0, 100.0))
+}
+
+/// Epoch-ms (MiniMax `end_time` / `weekly_end_time`) → ISO-8601 UTC.
+fn ms_to_iso(ms: i64) -> String {
```
uwuclxdy: Byte-identical fourth copy of this helper (zai's is private). Hoist one into providers/mod.rs.

uwuclxdy (changes requested):
The windows-to-chain change is ruled in: it ships as written. The direction matches what the bars always claimed, and the three limits (5h/7d only, `best_effort` never, own-cache reads) hold up against the chain docs. No hybrid guard is requested; see the recorded notes below.

Two things to change before merge, then finish work.

**The MiniMax in-band verdict breaks the dead-key contract.** `base_resp` is `#[serde(default)]`, so a 200 body without the envelope deserializes to success and publishes a healthy empty plan: cache overwritten, `Fresh`-stamped, and with part 2 the walk windows are removed. Every sibling fails closed on a missing verdict. And a rejected key's in-band `2049 invalid api key` maps to `Status` -> `Failed`, so a dead key never suppresses and never renders `(key rejected)`, which the wiki promises for any endpoint. Fail closed on the missing verdict and map auth-class codes to `AuthExpired` the way Alibaba's body verdict does.

**The hook-note nudge needs the same windows the walk now has.** `chain_would_act` replays with OAuth windows only, so the nudge stays silent where the live leg acts on a provider window. Extend the replay to third-party windows so both legs judge the same input.

More finish work:

- `edit_profile_endpoint` clears the in-memory stats on a provider change but never the disk cache, so the OLD provider's windows seed `Fresh` and drive the walk for up to one refresh interval after an endpoint edit. `remove_profile_cache(name, THIRD_PARTY_CACHE_FILE)` on the same change, plus a test.
- Interface-And-Keys.md's `open provider console` list omits MiniMax; the menu now offers it.
- Daemon.md's `windows[]` note is silent on the new source. Append: "On a third-party (api-key) profile the array carries that account's provider-derived 5h/7d windows, the two labels only. Never a `30d` or a best-effort scan's figure. `resets_at` is the provider's own reset instant, `null` when it reports none." And in the `third_party` note: "provider WINDOWS publish under `windows[]`, never inside this object". Shape stays additive, no schema bump.

The rest is inline: the console URL, the video row's window length, two Configuration.md lines, the mirror's removal-arm test, and a handful of nits.

Recorded, not requested: the hybrid double-writer on the shared store (both legs write one key per tick for an OAuth pair plus endpoint account; a bars-less hybrid loses its OAuth windows in the walk every fetch) ships as-is. The bell now fires for third-party accounts too; accepted without disclosure.

The wire shape checks out against two independent 2026 captures field for field, and the gate on this head is green: fmt, clippy, windows cross-lint, 3026 tests x 2 legs, audit. The Windows timing-sensitive note is filed on our side, unprobed.

uwuclxdy:
This is awesome btw, thanks!

uwuclxdy:
Follow-up on the Windows note: probed it on a real box, 3 parallel `cargo test --locked` runs of this head (2832 tests) and one of `v0.15.1` (2814), all green. The always-fails reading is ruled out ((1-0.99)^4 ~ 1e-8); 0-in-4 still leaves room for a low-rate flake, ~20% per run or less, and a different box may time differently.

uwuclxdy:
Landed everything from the review on your branch (`1c46807b..fbde4f2e`), nothing needed from your side.

- the in-band verdict now fails closed on a missing `base_resp`, and the dead-key codes map to `AuthExpired` so a rejected key stops polling and renders `(key rejected)`
- the hook-note nudge judges third-party windows like the walk does
- an endpoint or preset edit drops the stale third-party cache instead of reseeding the old provider's windows
- the console url points at the token-plan key page, and the `video` row states its real 24h window
- the wiki rows and a few stale comments caught up

Still curious what a wrong key actually answered in your live testing, if you kept the capture.

src/providers/minimax.rs
uwuclxdy: Landed: the envelope and `status_code` are non-defaulted (a missing verdict is a `Parse` error), and `1004`/`2049` map to `AuthExpired` so a rejected key stops polling and renders `(key rejected)`. The auth-class set was since verified against the vendor's error-code table.

src/providers/minimax.rs
uwuclxdy: Landed: the console URL points at `payment/token-plan`, the Subscription Key page, in code and the wiki row.

src/providers/minimax.rs
uwuclxdy: Landed: rows derive each bucket's window length from the response's own instants (`video` reads `24h`), and the separator is the single-spaced ` · `.

src/usage/scheduler.rs
uwuclxdy: Landed: the removal arm is pinned directly and through bootstrap, alongside the seed, the replay and both cache-drop paths.

wiki/Configuration.md
uwuclxdy: Landed: the section says eight and the table carries eight rows.

wiki/Configuration.md
uwuclxdy: Landed: the row now says "with more than one bucket and no `general`, no bars are drawn", matching the lone-bucket pin.

src/tui/app.rs
uwuclxdy: Landed: the hybrid precedence comment was rewritten to the store-entry-wins rule the mirror actually produces.

src/providers/mod.rs
uwuclxdy: Landed: both docs name all five providers now, including `ProfileEntry.windows` and the MCP render note.

src/profile_json.rs
uwuclxdy: Landed: renamed to `usage_windows` at the definition and every caller.

src/providers/minimax.rs
uwuclxdy: Landed: one `ms_to_iso` in `providers/mod.rs`, all three providers import it.

uwuclxdy:
Merged as `3e51210f`. All ten threads above landed and are resolved — thanks again for a well-flagged PR: the part-2 ruling ask and the three deliberate limits made the review easy.

The re-review round verified every fix against the code, ran the full gate on the trial merge into current `mommy` (which had moved ~90 commits through the same areas), and landed one more fix round on top: a seeded provider 5h window was taking the recency-weighted drain branch, which reads a history file no third-party leg appends, so the overview countdown lost its drain hue — the branch is now OAuth-only. The MiniMax auth-class set `{1004, 2049}` was also since verified against the vendor's own error-code table.

For the record: a wrong-key probe on `api.minimax.io` was never answered here; the `2049 invalid api key` verdict we mapped came from the round-1 capture, and the vendor table now confirms it. If you still have your live-testing capture, the HTTP status of that answer would still close the last open question.

PR #71, opened 2026-09-04
feat(login): manual login (no browser) for ssh hosts

wwilson1017:
> **operator asked:** "Is there currently a way to log into a Clauth account without using a browser? I want to be able to log in to an SSH session and still capture logins, like a headless server." The ask evolved during the session into: a second Setup-tab row named the way Claude Code names it (`manual login (no browser)`, with the existing row becoming `web login`), a matching `--manual` CLI flag, and a key that copies the link to the local clipboard over ssh.

Adds a browser-free OAuth login, `clauth login <name> --manual` and a `manual login (no browser)` row in the Setup tab, that mints the same credential the browser login does.

Claude Code's own `/login` builds two authorize URLs from one PKCE pair: the loopback one it opens in a browser, and a manual one whose redirect is `platform.claude.com/oauth/code/callback`, printed under "Browser didn't open? Visit:". That page shows a `code#state` string to paste back. clauth reproduced only the loopback flow, so on an ssh box the choices were a port forward, copying a profile dir from another machine, or a setup token (a different credential with no usage polling). I verified the manual redirect, the `code#state` shape, and the exchange body against the installed Claude Code binary (2.1.260); the token request clauth already builds is byte-for-byte the same, so no new network code was needed.

**Core (`src/oauth_login.rs`).** `MANUAL_REDIRECT_URI`; the tail of `login_with` after the code arrives is extracted into `finish_login` and shared. `begin_manual_login` returns a `PendingManualLogin` whose `parse` is non-consuming (a bad paste is corrected against the same link) and whose `complete` is consuming. `parse_manual_code` mirrors Claude Code's parser (split on the first `#`, both halves required) and adds the same `state` check the loopback callback makes; Claude Code itself leaves that to the server, so this is stricter on purpose. The pasted string is treated as a secret until exchanged: hand-written `Debug` on the pending login, the parsed code, and the TUI form; canned error text only; a 4 KiB cap on the trimmed code.

**CLI.** `--manual` prints the link, reads the code echo-off on a TTY (re-prompting after a malformed paste), or as one bounded line from a piped stdin for a driver feeding the same process. It is refused on an Alibaba Model Studio account, whose login is a console session, rather than silently running the console flow. Completions for bash, zsh, and fish learn the flag.

**TUI.** The manual row appears exactly where the login row would run the OAuth mint (never on an api-key or console account). Its modal has two phases so the copy key can never eat a pasted character: the link phase answers `c` (OSC 52 to the local terminal's clipboard, which crosses an ssh hop), enter, and esc; the code phase is a masked field where every printable character is data. A good paste seats a `LoginSession` at the exchange stage and joins the browser flow's worker, drain, and apply path unchanged. `LoginSession` carries its method so the progress modal and footer never say "complete it in your browser" for a manual exchange. On short terminals the link modal fits its own rows, reserving the key line first. The manual-row gate is a pure function of the profile `config_rows` already holds, because the config mutex is not reentrant.

**User-visible changes, per CONTRIBUTING:**
- `+ login` / `re-login` became `+ web login` / `web re-login`, on the OAuth mint only: a second login row exists now and the old label no longer said which flow it runs. Api-key and Model Studio accounts keep their labels.
- The OAuth login row's hint now reads "opens a browser on this machine; mints fresh tokens for this account", to contrast with the new row's hint.
- New row `+ manual login (no browser)` / `manual re-login (no browser)`, its hint, and its modal keys (`c`, enter, esc).
- The login footer says "manual login in progress" for a manual session instead of the browser line.
- The divergence-confirm detail says "a fresh login finished" instead of "a fresh browser login finished", since either flow reaches it.
- The `login` subcommand description in the zsh and fish completions mentions the manual code.
- No default values changed.

Docs updated where they described the login as browser-only: README command table, wiki Quickstart, FAQ, Interface-And-Keys, and SECURITY (network table and the "acts on your behalf" list).

Out of scope, found while testing on a fresh Ubuntu box and reproducible on `mommy` with the web login too: when Claude Code already holds its own login on a machine with zero clauth accounts, the first `create account` is refused ("live file differs … resolve the divergence in the TUI") because the divergence modal only exists for an active account, and the half-written profile dir is left behind. The README's "+ new from current profile" step it points at was removed in an earlier refactor. Filed as #72.

## agent

- Claude Code (claude-fable-5-1), operated by @wwilson1017
- ran: `cargo fmt --all -- --check`, `cargo clippy --locked --all-targets --all-features -- -D warnings` (debug and `--release`), `cargo nextest run --locked --all-targets --all-features --no-fail-fast` (debug: 3041 passed, 3 skipped; release: 3037 passed), `cargo build --locked --release`, all on macOS arm64; `cargo deny` and `cargo audit` are not installed here and the Linux/Windows legs are CI-only. Each of the two commits was also lint-checked on its own. New tests include an end-to-end exchange against a loopback stand-in asserting the token request body, and a keystroke-driven TUI run through the real worker. operator verified: the TUI manual flow on an Ubuntu 24.04 ssh box with no display (link opened on another device, code pasted, Max tokens minted twice); the subsequent account creation then hit the pre-existing first-account bug described above.
- unsure: whether platform.claude.com keeps accepting the manual redirect for this client id (it did in the operator's runs today); OSC 52 was exercised only in iTerm2; the state-mismatch hard error is stricter than Claude Code's own parser.


src/oauth_login.rs
```diff
@@ -616,6 +664,144 @@ pub(crate) fn login_with(
...
+        return Err(ManualCodeError::StateMismatch);
+    }
+    Ok(code.to_string())
+}
+
+/// Start a manual login: mint the PKCE pair and `state`, build the authorize
+/// URL against [`MANUAL_REDIRECT_URI`]. No network, no socket; the only failure
+/// is the CSPRNG. Callers show [`PendingManualLogin::url`] themselves — there is
+/// no [`LoginProgress::AuthorizeUrl`] event on this path.
+pub(crate) fn begin_manual_login() -> std::result::Result<PendingManualLogin, LoginError> {
```
uwuclxdy, line 763: This mints a fresh PKCE pair, which is the one thing that has to change for the one-flow shape: the manual URL should be built from the pair the in-flight loopback login already minted, the way CC 2.1.260 does it (this module's own doc comment says CC builds both URLs from one pair). With a fresh pair the `state` in a pasted code can never match the loopback session's, so the modal would be holding two pending logins. The fix shape: mint the pair up front and hand it to both URL builders, or expose `PendingManualLogin::from_parts` so the loopback flow can seed the manual one.

uwuclxdy (changes requested):
The core is in good shape and should stay: the pasted code gets the full secret treatment (redacted `Debug`, masked field, echo-off read, one bounded piped line, canned errors), the end-to-end test asserts the real token request body, and the phase split is the right way to keep `c` from eating a paste. What I'd like reshaped before this lands is the UX layer: one flow, not two.

The reason is the failure mode the two-flow shape leaves in place. `login_with` discards the `open_url` result (`let _ = ...`) and blocks on the loopback listener for up to 180s, so on a host with no browser the login modal offers `r` retry and no paste path, then dies on the timeout. That is exactly the case this feature exists for, and a user there has to have known in advance to pick the manual row instead. Claude Code ships the one-flow shape: loopback callback as the fast path, the manual link printed under "Browser didn't open? Visit:", a paste accepted alongside. I'd like the same here.

What I'd ask for:

- CLI: drop `--manual` and its completions. The bare login prints the hosted link under the existing fallback line and accepts a paste as an alternative to the loopback callback: echo-off TTY loop, one bounded piped line for a driver. Whichever lands first wins; the existing login generation guard already discards the loser.
- TUI: drop the `manual login (no browser)` row and the `+ web login` relabel. The login progress modal gains `c copy link` + `p paste code` below `r open the browser again`. `p` opens the existing code phase, seeded with the in-flight session's pair instead of a fresh `begin_manual_login`. Copying without rendering the 440-char link sidesteps the wrapping problem the link-phase modal's fit logic exists to solve.
- One PKCE pair for both URLs (inline note below), so a pasted code and the loopback callback are two doors into the same login.

What must survive the reshape: the hosted redirect plus the `code#state` parse and state check, the secret-handling set, the Alibaba and api-key carve-outs, the phase split, and `LoginMethod` on `LoginSession` so the footer copy stays honest during a paste exchange.

The one real argument for the two-flow shape is blast radius: it leaves the desktop login byte-identical. It doesn't win here. The default-flow change is small, and the separate shape is also the reason the 180s silent wait stays unfixed on the flow it doesn't touch.

uwuclxdy:
Thanks for the PR, the hosted-redirect half is what made the one-flow shape possible. Landed the reshape on your branch so it can merge as one login: `4f9c0ba9` merges current `mommy` in, `73634021` is the one-flow login (one PKCE pair builds both links, the browser callback and a pasted `code#state` race into `PendingLogin::run` and the exchange follows whichever door won; the CLI prints the hosted link under the fallback line and takes an echo-off paste beside the listener; the login modal gets `c copy link` and `p paste code`, the latter turning its own row into an inline code field), `dce90937` gives both CLI login flows one fallback printer, `1df59e9c` the wiki, `SECURITY.md` and README. `--manual`, the manual Setup row and the link modal went with it; the wait bound is 10 minutes now, since a phone round-trip rarely fits in three. Your core survived as it was: the parser and its state check, the redacted `Debug` impls, the bounded stdin read and the OSC 52 writer are all still yours.

PR #73, opened 2026-09-05
feat(selection): choose the account at session start by the models it will run

idatsy:
> **operator asked:** "Does clauth support using one account for opus and one for fable to better manage the 5 hour limits?" — which became "can that be done generically, selecting under the hood at each new session start (including subagents), and shaped as an upstream contribution rather than a personal patch". The ask evolved: an initial "never switch mid-session" invariant was relaxed once we agreed the chain is the right rescue when an account genuinely runs out.

Picks the account a session starts on from the models that session may run, so an account capped only on a per-model week it will never touch stops being excluded from work it could do.

The chain decides where a session *moves*; nothing decided where it *starts*. That is why a `weekly_scoped` window has to be a blanket exclusion — `scoped_weekly_blocked_info` says it outright: the walk cannot know which model the next session runs. `check_scoped = false` is the only escape and it is all-or-nothing, so a session of the capped model can then be routed onto a member that cannot serve it. A session that has not started yet is the one case where the model *is* knowable, which is the gap this fills.

New `src/selection.rs` holds the selector. `clauth start --auto` weighs the chain and starts on the best member that can serve every model family the session may run. `worst_scoped_window` gains a `_for` variant taking those families; the blanket spelling delegates to it, so the narrowed judgment cannot drift from the chain's — the liveness rule in particular, without which an already-reset window would gate off a stale cache. I got that wrong on the first pass by inlining the comparison, which is what the "never a second opinion" comment is there to prevent.

Selection is on the **union** of families rather than the headline model. A `Task` subagent runs inside the parent process and spends the parent's account on whatever model it runs, so choosing for the main thread alone would strand the session the moment a subagent used a capped family. The union comes from the effective `settings.json` model, `CLAUDE_CODE_SUBAGENT_MODEL`, and any `--model` in the passthrough. With none resolved the demand is empty and the existing blanket `check_scoped` behaviour stands.

Feasibility is **runway** (headroom over burn, reusing `compute_burn_rates_from_history`), not utilization. Picking a member with minutes left causes precisely the cost start-time selection avoids: the session strands one turn in and the chain swaps it, re-reading the whole context because cache entries do not cross accounts. Absent burn reads as unbounded rather than zero, matching `is_exhausted_projected`'s stance that a member is never uncovered for lack of data — an idle account has no samples *because* it is idle. A window resetting inside the grace is feasible however thin. Nothing feasible still launches, flagged, rather than refusing.

Ranking puts live sessions ahead of runway: launching several sessions at once would otherwise send them all to the same member, and usage polling cannot see the launches until its next refresh. A member with no usage cache ranks below one you have numbers for — it reports unbounded runway for want of data, not because it is empty, and an api-key chain member would otherwise win every selection. I only caught that on a re-read, not from a failing test.

Two smaller things ride along. `delegate` can pick its own account when `profiles` is omitted (opt-in, `delegate_auto_select`, default off) — a separate process, so the one subagent that can hold one. And `clauth start --explain` prints the decision without spending a window on it, since every other decision surface here renders its reason.

Nothing existing changes behaviour. `--with-fallback` is untouched: this never moves a running session, and the two compose. `clauth start <profile>` and the chain walk are identical in effect, and all three new config keys default to today's semantics.

**User-visible surfaces**, per CONTRIBUTING:
- new `start` flags `--auto` and `--explain`, added to all three completion scripts. `--auto` deliberately does *not* complete a profile after it — it takes the profile's place, so what follows belongs to `claude`. The obvious edit (adding it beside `--with-fallback` in the profile arm) reads as consistent and is wrong; there is a test pinning that.
- `--auto` requires the `--` separator when `claude`'s first argument starts with a hyphen. With no name in the profile slot clap cannot tell a passthrough `-p` from a misspelled clauth flag. I chose the separator over giving the positional `allow_hyphen_values`, which would turn `clauth start --isolatd acme` from a clean flag error into a confusing profile-not-found. This matches the convention `start --help` already documents, and three tests pin it.
- new `profiles.toml` keys: `selection_min_runway_mins` (20), `selection_reset_grace_mins` (10), `delegate_auto_select` (false).

Docs are in the same commit: Quickstart (flags table + two rules), Configuration (the key table), Auto-Switch (the "cannot know which model" passage, the `scoped gate` bullet, and a new *Choosing where a session starts* section), Claude-Code-Plugin (delegate targets), FAQ, README. No TUI surface for the new keys yet — they are file-only, and I did not want to guess at Config-tab layout.

## agent

- Claude Code (claude-opus-5), operated by @idatsy
- ran: no `cargo.sh` in the tree, so the three it wraps. `cargo fmt --all -- --check` clean. `cargo clippy --locked --all-targets --all-features --release -- -D warnings` reports one pre-existing `nonminimal_bool` in `src/tui/render/footer.rs:317`, a file this PR does not touch; it reproduces on an unmodified `origin/mommy` worktree with clippy 0.1.96, and your CI was green on that commit, so it is a newer-toolchain hit rather than something I introduced — I left it alone rather than widen this PR. `cargo test` 3050 passed / 2 failed, both `mode 0644 vs 0664` assertions in `claude_json` and `settings_sync` that fail identically on unmodified `origin/mommy`; my box's umask is 0002. I checked both in a throwaway worktree rather than assume. `every_completion_script_parses_in_its_own_shell` skips silently without zsh/fish installed, so I ran `bash -n`, `zsh -n` and `fish --no-execute` over the three generated scripts in a container instead — all parse. Also dry-ran selection against a credential-free copy of a real four-account fleet: with a member at a per-model week of 100% and an idle 5h window, an opus launch selects it and a fable launch correctly does not. Operator verified: reviewed the diff and the payload before this was opened; has not run the branch themselves.
- unsure: the ranking weights are my judgement, not measured — fewest-live-sessions ahead of most-runway is the ordering I would most expect you to want changed, and the 20/10 minute defaults are guesses that felt safe rather than tuned. I chose the fallback chain as the candidate set rather than all enabled profiles, reasoning that the chain is the operator's own statement of which accounts may be entered unattended; an empty chain refuses with the fix named. `delegate_auto_select` has unit coverage for the selector but no test against a live delegate spawn. And `scoped_label_is` matches any word in the label, which is my guess at how a multi-word model display name would arrive — I have only ever seen single-word ones.


uwuclxdy:
Thanks for the PR, this is a real gap: the walk excludes a scoped-capped member as a blanket because it can't know which model the next session runs, and `check_scoped = false` is the only escape. The narrowing (`worst_scoped_window_for`) is exactly right, liveness handling included, and I'd like that half to land.

What I'd re-pose is the selector built on top of it. The shipped shape is an 883-line ranking policy with hardcoded weights that overrides the chain order you yourself call the operator's statement, plus three file-only config keys with no TUI row, on a feature whose value is efficiency (a strand is already rescued by `--with-fallback`, at the cost of re-sending context). That reads as an optimization-sized ship on a safety-sized gap.

The reshape I'd ask for: pass the families into the existing walk's headroom-accept predicate instead of a new selector. `worst_scoped_window_for` already makes that a one-parameter change: first-acceptable along the chain order, no new ranking, no new keys, no new module. `--explain` stays, rendering the walk's reason the way every other decision surface here does. The `delegate_auto_select` arm is defensible, but it is a second selection surface, doubles the review scope, and doesn't strengthen the start-time case: split it out into its own proposal.

Findings that survive any reshape, because they are claims about the pieces I'm asking you to keep:

- `--explain` prints the dry-run line before `start::run`'s `--with-fallback` refusals (non-OAuth, not a chain member, a singleton chain, no daemon), so it can name a target the real launch rejects. The comment above it says it answers "what a real start would do". The explain path should run the same refusal set the launch runs.
- "reads live usage" is stronger than the code: selection reads `usage_cache.json` with no freshness or age check, so a days-old cache with no daemon running is judged as current and disclosed as nothing. Either gate on freshness or print the cache's age.
- `model_family` drops every claude-3.x id (the digit guard at `selection.rs:100`), so a mixed union silently omits a family. A family matcher must not silently narrow the demand.

The rest of the findings live in the dropped half and go away with it, so they are not asks: the 5h runway arm has no liveness gate, `kick_block.json` is never read, the runway floor degenerates to the 5h window in practice (weekly/scoped burn practically never slopes), the runway-arm demand filter and the reset-grace equality boundary are unpinned (I planted a label-blind filter and all 22 selection tests stayed green), and `--auto --with-fallback` composes on paper but the selector doesn't filter OAuth-ness, chain size, or daemon liveness, so the launch refuses the pick.

Gate on the trial merge of `d1cd94c` onto current `mommy` (four mechanical conflict resolutions, 147 commits of drift): locked check, clippy, fmt, windows cross-lint, nextest release and debug, deny, and audit all green.

src/cli.rs
```diff
@@ -351,8 +351,29 @@ pub(crate) struct StartArgs {
...
+    /// It takes the place of the profile name, so separate `claude`'s own args
+    /// with `--` whenever the first of them starts with a hyphen:
+    /// `clauth start --auto -- -p "hi"`. Without a name in that slot there is
+    /// nothing to tell a passthrough `-p` from a misspelled clauth flag, and
+    /// guessing would silently eat one of them.
+    #[arg(long)]
+    pub(crate) auto: bool,
+    /// Print the account that would be launched, and why, without launching it.
+    ///
+    /// The selector reads live usage and burn history, so this is how an
```
uwuclxdy, line 370: "reads live usage" is stronger than the code: the selector reads `usage_cache.json` with no freshness or age check, so a days-old cache with no daemon running is judged as current and disclosed as nothing. Gate on freshness, or print the cache's age in `--explain`.

src/main.rs
```diff
@@ -296,15 +303,65 @@ fn cmd_completions(target: &str, shell: Option<&str>) -> Result<()> {
...
     let config = load_config()?;
-    let canonical = resolve_or_bail(&config, name)?;
+    let canonical = match target {
+        cli::StartTarget::Named(name) => resolve_or_bail(&config, name)?,
+        cli::StartTarget::Auto => auto_select(&config, rest)?,
+    };
     refuse_if_disabled(&config, &canonical)?;
+    // After the refusals, so `--explain` answers what a real start would do
+    // rather than naming a target the launch would then reject.
+    if explain_only {
```
uwuclxdy, line 323: The dry-run returns here, before `start::run`'s `--with-fallback` refusals (non-OAuth, not a chain member, a singleton chain, no daemon), so `--explain` can name a target the real launch rejects. The comment above this line says it answers what a real start would do. Run the same refusal set the launch runs before printing.

src/selection.rs
```diff
@@ -0,0 +1,883 @@
...
+    }
+    // Bare aliases as Claude Code spells them in `settings.json` and `--model`.
+    // `opusplan` is deliberately absent: it is two families and resolves in
+    // `demand_from`, which can return both.
+    if matches!(id.as_str(), "opus" | "sonnet" | "haiku" | "fable") {
+        return Some(id);
+    }
+    let rest = id.strip_prefix("claude-")?;
+    let family = rest.split(['-', '.', '[']).next()?;
+    if family.is_empty() || family.chars().any(|c| c.is_ascii_digit()) {
```
uwuclxdy, line 100: The digit guard drops every claude-3.x id (`claude-3-5-haiku`, `claude-3-7-sonnet`) from the demand union, so a launch mixing a parseable family with a 3.x id silently omits a family. A family matcher must not silently narrow: skip a leading digit component and take the first alpha token.

uwuclxdy:
Thanks for the PR, the narrowing is what made the fold possible. Landed the reshape on your branch so it can merge as one change: `b5d69eca` merges current `mommy` in, `2619623c` folds the selection into the chain walk (the families narrow `worst_scoped_window_for` for every member in chain order, and a known model outranks the `scoped gate` toggle; `--auto` is the walk's first clear member, a dated reading preferred over a stale or missing one; `--explain` runs `start::admit`, the launch's own refusals, then prints every member's verdict in the Fallback tab's words with the age of its usage reading; `model_family` keeps the 3.x ids and the `[1m]` aliases, `best` counts as fable + opus, and the fallback chain from `--fallback-model` / `fallbackModel` joins the union), `cceddafd` the wiki. The ranker, its three keys, `selection.rs` and the delegate arm went with it; the delegate pick is still yours to propose on its own. Your `worst_scoped_window_for`, `claude_settings_models`, the `--` handling for `--auto` and the CLI tests survived as they were.

PR #77, opened 2026-09-16
feat(profile): choose the home account by weekday

kotahashihama:
> **operator asked:** "I'd like clauth to handle day-of-week too, so a PR would help" — he runs a work account and a personal one and wants a different one to be home on weekends. Translated from Japanese.

Adds a per-profile `preferred_days` list: the weekdays that account is the chain's home account, read in the machine's local zone.

```toml
# ~/.clauth/profiles/personal/config.toml — work keeps its plain `preferred = true`
preferred_days = ["sat", "sun"]
```

**A named day is claimed against every account.** On a day some list names, only the accounts naming it are home; a bare `preferred = true` elsewhere stands down for that day and takes charge again on the days no list claims. So the split above is one line in one profile, and any weekday can go to any account. An install with no list anywhere behaves exactly as before.

Resolving this across the profile list rather than per profile is the whole point, and it is the one design call I would most like checked. Per profile, the flag would keep claiming all seven days, so on a Saturday both accounts would read as home and `chain.iter().find(|m| m.preferred)` would settle it by chain order — an operator who wrote one line would get a silent coin flip. `AppConfig::is_home_on` answers it in one place instead.

Entries parse through chrono's `Weekday`, so full names and three-letter forms work in any case. One that does not parse is dropped instead of failing the profile, and `maybe_rewrite_config_toml` then drops it from disk, which is the only signal an operator gets that it was not understood. Rendering is canonical lowercase three-letter names, so a hand-written `["Saturday", "SUN"]` settles instead of alternating on every rewrite. `duplicate_profile` leaves the list behind for the same reason it leaves `preferred`, which its doc comment now says.

The day is read in `chain_member` (`src/fallback.rs`) and `scan_recovery` (`src/usage/scheduler.rs`) — where a `ChainMember` is built — rather than at load. `ReloadFingerprint` is built from `config.toml` mtimes, and midnight moves no file, so a load-time fold would sit stale until the next config write. Per chain build it costs one `Local::now()` on a path that already runs each tick. The `preferred` field is untouched: the TUI toggle, the radio against `last_resort`, and the return-to-preferred pass read what they read today.

**Two user-visible strings move, both because they would otherwise describe a home the chain would not pick** (flagging these per CONTRIBUTING):

- the Fallback card's `preferred` hint reads the lists on both sides of one: it names the days when this profile has a list, and says "on the days no day list claims" when another profile does. The stock wording is a promise the chain would not keep on the days it stands down.
- the Overview's `⌂` — the glyph marking a switch that lands on home — reads the same resolved value the chain does. It read the raw flag before, so on a weekend it marked a hop onto the weekday account.

Two accounts naming the same day is not rejected; the earlier chain member wins it, the same first-match a hand-edited double `preferred = true` already hits. The wiki says to keep the lists disjoint rather than adding a validation path, since that invariant is enforced by the TUI toggle and not on load.

Tests are deterministic — `is_home_on` takes the day, and only `is_home_today` reads the clock:

- `tests/inline/profile.rs`: the empty default, parsing (case, short forms, duplicates, one typo), a render-to-parse round-trip, an unclaimed day falling back to the flag, and a claimed day standing the flag down on the other account.
- `tests/inline/fallback.rs`: the wiring into `ChainMember`, which the config-level tests cannot reach. A list naming every weekday resolves true whatever day the suite runs on; an empty list leaves the flag in charge.
- `tests/inline/tui_render_chain.rs`: the hint names the days.

Three commits, each `cargo check --all-targets` clean on its own: the feature, the two TUI surfaces, and the wiki.

## agent

Claude Code (claude-opus-5), operated by @kotahashihama.

ran: `cargo fmt --all -- --check` clean. `cargo clippy --locked --all-targets --all-features --release -- -D warnings` fails on `src/tui/render/footer.rs:329` (`nonminimal_bool`) — the same command on `mommy` fails identically, so it is rustc 1.94.1 here being stricter than what your CI resolves to. `cargo test`: 3726 passed, 1 failed, on all three runs — always `daemon::api::tests::a_silent_client_is_dropped_without_holding_a_slot_for_the_whole_lifetime`. I ran the suite on `mommy` three times to place it: one clean run, then the same test failing twice. It asserts a wall-clock deadline under whatever load the suite puts on the box, so it is flaky here rather than something this branch moved; it also passes on its own on both sides. operator verification: design only, not line by line.

unsure: the key name, and leaving two accounts that claim the same day to first-match instead of validating them. Also whether this wants a TUI editor rather than a hand-edited list — `preferred` and `last_resort` are both toggles on the Fallback card, and this one only renders.


uwuclxdy:
Requesting changes, but the core is right: exclusive day claims resolved at chain build, one line per profile, the flag governing unclaimed days. That design stands, and the trial merge gates green on `mommy` (fmt, clippy `-D warnings`, 3932 tests, windows cross-lint, deny/audit). The two local failures in the body don't reproduce here and aren't blockers.

Beyond the inline threads:

- **claim scope**: a day list must claim only from chain members that aren't dead. As written `is_home_on` scans the whole roster, so an account removed from the chain, or a disabled / auth-broken one, still stands the flag down on its days and leaves nobody home. The chain-behaviour precedent is the spend warning: reads over the chain, never `config.profiles`, because a non-member could never actually serve. The editor follows: show all accounts in the picker with an error toast with why a dead account cannot be added, or an error when a dead account already in the chain claims a day.
- **double claim**: first non-dead lister wins, not bare first-match. Today a broken first lister bails the whole return pass for the day (`fallback.rs:1803-1812`, the `find` runs before the skip gate). Use the next lister instead, with a one-shot warning log + toast. The pass re-runs every tick, so the warning needs a once-per-day (or per-config-change) gate.
- **the three `returned to preferred` surfaces still key on the raw flag**: `daemon/tick.rs:280`, `app.rs:2585`, `app.rs:10093`. On a claimed day they describe the destination as not-home while the `⌂` says home, the same defect this PR moved two strings to fix. Point them at `is_home_today` like the glyph.
- **`ChainMember.preferred`'s doc** (`fallback.rs:881-885`) is now false twice: it doesn't mirror the flag anymore, and "at most one member carries it" breaks on a claimed day with two listers.
- **`--with-fallback` help** (`cli.rs:440`) names only the flag as the home marker; the list is the second marker now.
- **TUI editor**: the list wants an editor on the Setup tab, the account's own card. A day row there; the Fallback card keeps its hint-only mention and the wiki stops calling it hand-edited-only.

src/profile.rs
```diff
@@ -1092,6 +1101,39 @@ impl AppConfig {
...
+    /// only the profiles naming it are home, and a bare `preferred = true`
+    /// elsewhere stands down for that day. Resolving this per profile instead
+    /// would leave the flag claiming all seven, so the two-account split an
+    /// operator actually wants — weekdays here, weekends there — would need a
+    /// list on both sides, and getting one wrong reads as first-match luck in
+    /// `fallback.rs` rather than as a mistake.
+    ///
+    /// On a day nobody names, `preferred` decides exactly as before.
+    pub(crate) fn is_home_on(&self, name: &ProfileName, day: Weekday) -> bool {
+        let claimed = self
```
uwuclxdy, line 1117: major: the claim scan reads the whole roster, so a profile removed from the chain, or a disabled / auth-broken one, still stands the flag down on its days and leaves nobody home. Restrict the claimed check to chain members that pass `walk_excluded` and pin both cases with tests (a non-member's list reads inert, a dead member's list hands its days back to the flag).

src/tui/render/chain.rs
```diff
@@ -766,8 +766,32 @@ fn last_resort_hint(cfg: &AppConfig, name: &crate::profile::ProfileName, on: boo
...
+        .find(name)
+        .map(|p| p.preferred_days.clone())
+        .filter(|d| !d.is_empty())
+    {
+        let named = days
+            .iter()
+            .map(|d| d.to_string().to_ascii_lowercase())
+            .collect::<Vec<_>>()
+            .join(", ");
+        return format!("home on {named}, by the day list in this profile's config.toml");
```
uwuclxdy, line 783: major: `on` is never read. With a list AND the flag on one account, the flag still decides the unclaimed days, so that account is home all seven and this hint says only the listed ones. Make the branch value-aware (list days plus what the flag holds on the rest), the way the `last resort` twin reads.

src/tui/render/chain.rs
```diff
@@ -766,8 +766,32 @@ fn last_resort_hint(cfg: &AppConfig, name: &crate::profile::ProfileName, on: boo
...
+    // here the toggle answers nothing; with a list elsewhere the toggle still
+    // holds, but only on the days that list leaves alone.
+    if let Some(days) = cfg
+        .find(name)
+        .map(|p| p.preferred_days.clone())
+        .filter(|d| !d.is_empty())
+    {
+        let named = days
+            .iter()
+            .map(|d| d.to_string().to_ascii_lowercase())
```
uwuclxdy, line 780: minor: `profile` reads wrong in display copy, the house copy says `account`, and this lowercase map mirrors `render_preferred_days` in `profile.rs`. Make that helper `pub(crate)` and call it.

src/profile.rs
```diff
@@ -3060,6 +3136,30 @@ fn render_config_toml(profile: &Profile) -> String {
     }
     out.push('\n');
 
+    out.push_str("# Weekdays this profile is the home account, in local time. Empty (the\n");
+    out.push_str("# default) leaves `preferred` above in charge every day; a non-empty list\n");
+    out.push_str("# replaces it for this profile, so one account can own the work week and\n");
```
uwuclxdy, line 3141: minor: the list doesn't replace the flag, on days no list claims `preferred` still decides for this profile too. The wiki table row says the same; word it as claiming these days while the flag holds the rest.

tests/inline/tui_render_chain.rs
```diff
@@ -251,6 +252,39 @@ fn last_resort_hint_names_the_currently_marked_member() {
     assert!(hint.contains("instead of 'b'"), "{hint}");
 }
 
+// A day list outranks the toggle, so the hint has to name the days rather than
+// the toggle's state: read the stock wording with a list set and an operator
+// would take the toggle for the answer.
+#[test]
+fn preferred_hint_names_the_day_list_when_one_is_set() {
```
uwuclxdy, line 259: minor: only the list-set branch is pinned; the `claimed_elsewhere` branch ("on the days no day list claims") renders with no test. One more render pin for it.

wiki/Auto-Switch.md
```diff
@@ -62,6 +62,8 @@ Two radio toggles on a member's Fallback card. Marking one clears it on every ot
 - **`last resort`** is the parking spot: chosen only once every other member is past its line, and never switched away from. Claude Code then surfaces its own out-of-limit message when that account runs dry too.
 - **`preferred`** is the home account. Once it reads clear and fresh, clauth walks back to it on its own, from wherever the chain left you.
 
+Which account is home can also depend on the day. A `preferred_days` list in a profile's `config.toml` ([Configuration](Configuration#configtoml)) names the weekdays that account is home, in local time, and those days are claimed against every account: a plain `preferred` elsewhere stands down on them and holds the rest. One line on the weekend account is the whole weekday / weekend split, and the rollover needs no restart.
```
uwuclxdy, line 65: nit: the day list only reaches the claude chain, the codex chain has no home concept, so `preferred_days` does nothing there. One clause saying claude-only keeps a codex reader from expecting it.

src/profile.rs
kotahashihama: df656ad. `is_home_on` reads a new `day_listers`, which filters `fallback_chain` through `walk_excluded`, so only a member the walk would actually visit can claim. Both cases are pinned: a non-member's list reads inert, and a disabled member's days go back to the flag.

src/tui/render/chain.rs
kotahashihama: 47d40f0. The list branch takes `on` now. With both set it reads "home on sat, sun by the day list, and on the rest by this toggle"; with the flag off it names the list alone.

src/tui/render/chain.rs
kotahashihama: 47d40f0. Says `account`, and calls `profile::render_preferred_days` — now `pub(crate)` — instead of mapping the days a second time.

src/profile.rs
kotahashihama: df656ad for the config.toml comment, 8a26f53 for the wiki row. Both say the list claims its days against every account while `preferred` keeps the days no list claims.

tests/inline/tui_render_chain.rs
kotahashihama: 47d40f0. The `claimed_elsewhere` branch has its own render pin now.

wiki/Auto-Switch.md
kotahashihama: 8a26f53. The paragraph closes on the list reaching the claude chain alone, since the codex chain has no home account.

kotahashihama:
All six are in.

- **claim scope** — df656ad. `is_home_on` reads a new `day_listers`, which walks `fallback_chain` through `walk_excluded`, so an off-chain or dead lister hands its days back to the flag instead of standing it down. Pinned both ways.
- **double claim** — 4b50fbe for the walk, 65b8ccc for the warning. The return pass and `find_recovered_member` both `filter().find()` over every home member now, so a broken first lister no longer bails the day. For the warning I used the notice text itself as the gate key rather than a bool: it names the day and the claimants in chain order, so it changes at the rollover and on a config edit and is byte-equal in between. A bool would also stay silent if a second list arrived while the first collision was still up. The TUI toasts and logs; the daemon logs alone.
- **the three return surfaces** — fd1cc60. `daemon/tick.rs:280`, `app.rs:2585` and `app.rs:10093` read `is_home_today` like the glyph.
- **`ChainMember.preferred` doc** — 4b50fbe. It now says the value is resolved by `is_home_today`, and that every lister carries it on a claimed day.
- **`--with-fallback` help** — 8a26f53.
- **TUI editor** — 80d944f, with 21a6531 dropping the wiki's hand-edit line. A `home days` row sits beside `auto-start` on the account's own card; the Fallback card keeps its hint-only mention.

On the editor and a dead account: a list on one that is off the chain, disabled or auth-broken is saved and then explained, naming the first blocker in the way, and the row's hint carries the same reason before anything is typed. I did not refuse the save — that would hide a state `is_home_on` already handles, and leave no surface for an account the operator is about to put back on the chain.

User-visible additions, per CONTRIBUTING: the `home days` row and its hint; a refusal toast naming an entry that is not a weekday; a warning toast for a list that claims nothing; and the collision toast above.

On the two local failures — I could not reproduce them either. A second full run failed a different pair of `watchdog` tests instead, so I am reading all of them as timing under parallel load rather than as anything this branch does. Each passes on its own.

src/profile.rs
```diff
@@ -1173,6 +1183,87 @@ impl AppConfig {
...
+    /// off the chain, or one `walk_excluded` skips (unresolvable, auth-broken,
+    /// disabled), never serves, so letting its list stand the flag down would
+    /// leave the day with nobody home. Reading the chain rather than
+    /// `profiles` follows the spend warning, which is on the chain for the
+    /// same reason.
+    pub(crate) fn is_home_on(&self, name: &ProfileName, day: Weekday) -> bool {
+        let claimed = self.day_listers(day).next().is_some();
+        self.find(name).is_some_and(|p| {
+            if claimed {
+                p.preferred_days.contains(&day) && !crate::fallback::walk_excluded(self, name)
```
uwuclxdy: minor: this branch gates on `walk_excluded` but not on chain membership, so an off-chain healthy account reads as home on a claimed day while the lister scan refuses the same claim. gate it on membership too, matching the predicate's own doc.

uwuclxdy:
Reviewed the new head (`8a6caa81`, the mommy merge included) — every round-1 ask verified in code, and the conflicts are gone with it. The walk-order interplay holds in the merged tree: the preferred pass keeps chain order and outranks the recomputed soonest-reset home (`fallback.rs` `!any(preferred)` guard), `find_recovered_member` orders preferred before `ordered_walk`, and `scan_recovery` carries both. Gate: the full gate is green on the reviewer's trial resolution of the same merge (4090 + 4094 tests, windows cross-lint, deny, audit); the targeted walk-order and day-list pins also pass on your merged head.

Two one-line fixes, inline below: the claim predicate answers wrong for a non-member, and the own-list hint branch ignores eligibility while `claimed_elsewhere` does not.

The ruling-3 ask still stands: a dead lister passed over raises no tick-time notice. The editor-side hint + save warning stay; add the once-gated daemon logline + TUI toast for a day whose first claimant is dead and whose second carries the day, gated on the same message-key scheme.

Ruling 2 is settled in your favor: the save-with-warning stands, `is_home_on` already refuses the claim at walk time.

The nits (unpinned logline, the half-false picker comment, the toast tail on an off-chain account, the redundant allow) can ride the same push.

src/tui/render/chain.rs
```diff
@@ -766,8 +766,38 @@ fn last_resort_hint(cfg: &AppConfig, name: &crate::profile::ProfileName, on: boo
...
+        .find(name)
+        .map(|p| p.preferred_days.clone())
+        .filter(|d| !d.is_empty())
+    {
+        let named = crate::profile::render_preferred_days(&days).join(", ");
+        return if on {
+            // 80 columns is the narrow case this card renders at; the tail has
+            // to survive it or the operator reads only the list half.
+            format!("home on {named} by the day list, and on the rest by this toggle")
+        } else {
```
uwuclxdy: minor: this branch reads the list without the eligibility check, so a disabled member's card promises home on days `is_home_on` refuses. gate it like the `claimed_elsewhere` branch below.

src/daemon/tick.rs
```diff
@@ -51,6 +51,7 @@ impl super::Daemon {
         // window or this deadline is spent.
         let deadline = Instant::now() + super::WATCHDOG_DEADLINE.saturating_sub(super::TICK);
         self.reload_if_changed();
+        self.log_day_claim_collision();
```
uwuclxdy: nit: the collision notice's logline half is asserted by no test, only the once-gate transitions. consider pinning the emission through the `LogLines::capture_here` seam.

src/tui/app.rs
```diff
@@ -7963,6 +8000,80 @@ fn commit_config_field(app: &mut App, field: ConfigRow) {
...
+/// from the saved value so the canonical spelling lands in the field.
+///
+/// An entry that does not parse leaves the editor OPEN with the typing intact
+/// — the loader drops a bad entry because a file nobody is watching must still
+/// load, but here the operator is standing at the field and can fix it.
+///
+/// A saved list on an account the chain walk would skip claims nothing
+/// (`AppConfig::is_home_on` lets only members past `walk_excluded` claim), so
+/// the save is followed by the reason. Saved rather than refused, and shown
+/// rather than hidden, for the reason the chain picker lists dead accounts:
```
uwuclxdy: nit: the cited precedent is half-false, `chain_candidates` filters disabled accounts so the picker does not list them. reword the comment to name what actually gets listed.

src/tui/app.rs
```diff
@@ -7963,6 +8000,80 @@ fn commit_config_field(app: &mut App, field: ConfigRow) {
...
+                    *input = InputState::new(&value);
+                }
+                d.active = None;
+            }
+            if let Some(reason) = blocker {
+                app.toast(
+                    ToastKind::Warning,
+                    format!(
+                        "saved, but this list claims nothing: {reason}\nthe `preferred` toggle \
+                         still decides those days"
```
uwuclxdy: nit: on an off-chain account the toggle this tail promises is equally chain-inert, so nobody decides those days. reword to name what actually governs.

tests/inline/daemon_mod.rs
```diff
@@ -1549,3 +1549,51 @@ fn a_daemonless_publish_yields_to_a_feed_written_after_its_build_started() {
...
+
+    daemon.log_day_claim_collision();
+    assert_eq!(
+        daemon.day_claim_notice.as_deref(),
+        Some(first.as_str()),
+        "an unchanged tick leaves the gate where it was"
+    );
+
+    {
+        #[allow(clippy::expect_used)]
```
uwuclxdy: nit: redundant allow, the file opens with the per-file `expect_used` allow.

uwuclxdy:
The full gate on your merged head is green end to end: 4090 + 4094 tests, windows cross-lint, deny, audit.

src/profile.rs
kotahashihama: aabc059. The branch asks `day_listers` instead of re-deriving its predicate — home on a claimed day IS the claimant set — so an off-chain account cannot answer differently from the scan. Pinned with a fixture that also puts a chain member on the day, which is what the existing non-member test was missing.

src/tui/render/chain.rs
kotahashihama: aabc059. Gated on `day_claim_blocker`, so an inert list falls through to the toggle wording rather than promising days the scan refuses. The reason is left to the Setup tab's `home days` row — this card cannot act on it.

src/daemon/tick.rs
kotahashihama: 0fe2650. The test takes `LogLines::capture_here` and asserts on the lines, so dropping the `logline!` reds it. The gate transitions stay, since they answer a different question.

src/tui/app.rs
kotahashihama: 303b583. Dropped the picker citation. The reason that does hold is that the state is reachable without this row at all — a list goes inert when the account later leaves the chain or its login breaks — so the comment says that instead.

src/tui/app.rs
kotahashihama: 303b583. Now "the chain decides those days without this account", which is what actually happens. The old tail named a toggle that is equally chain-inert here.

tests/inline/daemon_mod.rs
kotahashihama: 0fe2650. Dropped.

kotahashihama:
Both one-liners and ruling-3 are in, plus the four nits.

- **claim predicate** — aabc059. `is_home_on` asks `day_listers` now instead of re-deriving its eligibility, so the two cannot answer differently at all. Worth noting the existing non-member pin passed through the old code: its fixture had no chain lister, so `claimed` was false and the flag branch answered. The new pin puts a member on the day as well, and fails against the old predicate.
- **own-list hint** — aabc059. Takes the check `claimed_elsewhere` already applied to the other side and falls through to the toggle wording. The reason stays on the `home days` row, which is the surface that can act on it.
- **ruling-3** — 0fe2650. Once-gated logline + toast, same message-key scheme. I widened it a little past the carried case: the notice fires for any lister that cannot serve today, and the tail says what became of the day: another list carried it, or nothing did and it fell back to the flag. The uncarried case is the same defect (a line doing nothing) and reads worse, since the day has silently gone back to the flag. Say the word if you want it narrowed to the carried case.
- **unpinned logline** — 0fe2650, through `LogLines::capture_here`, so dropping the `logline!` reds the test. Both gates hold the SET of notices now rather than one string: two notices are two keys, and one string would either repaint the first or swallow the second.
- **picker comment** and **toast tail** — 303b583. **redundant allow** — 0fe2650.

One thing I did not touch: `is_home_on`'s unclaimed branch still reads the bare flag with no eligibility check, so a disabled chain member with `preferred = true` reads home on a day no list claims. That is byte-for-byte what `chain_member` did before this branch, so it is not a regression here — but it is the same shape as the claim-side gap you just found, and I did not want to widen the PR into it without asking.

`cargo test` on the pushed head: 4115 passed, 0 failed. `mommy` has moved four more commits since; they merge clean, so I have left the branch where it is.

uwuclxdy:
The widening stands: the inert-list case is the same defect and reads worse, keep it.

The unclaimed branch is in scope too: a disabled member with `preferred = true` reads home on a day no list claims, the same shape as the claim side you just fixed. gate it with the same eligibility the scan applies and pin the disabled-preferred case.

src/profile.rs
uwuclxdy: the claim scan reads the chain now: `day_listers` over `fallback_chain` filtered by `walk_excluded`

src/tui/render/chain.rs
uwuclxdy: the branch is value-aware now and both arms pinned

src/tui/render/chain.rs
uwuclxdy: `account` wording landed

src/profile.rs
uwuclxdy: the config comment and wiki row reworded to claim semantics

tests/inline/tui_render_chain.rs
uwuclxdy: pinned: `preferred_hint_says_which_days_are_left_when_another_account_claims`

wiki/Auto-Switch.md
uwuclxdy: the Auto-Switch clause landed

src/profile.rs
uwuclxdy: the branch IS the scan now (`day_listers`), no second predicate

src/tui/render/chain.rs
uwuclxdy: gated on `day_claim_blocker` like the other side

src/daemon/tick.rs
uwuclxdy: pinned through `LogLines::capture_here`, set-of-notices keys

src/tui/app.rs
uwuclxdy: comment reworded

src/tui/app.rs
uwuclxdy: toast tail reworded to name the chain

tests/inline/daemon_mod.rs
uwuclxdy: allow dropped

uwuclxdy:
The entry guard landed and pins the disabled-preferred case; merged.

uwuclxdy:
image
image

this setting will be under the fallback tab. use space, enter and arrow keys to change

Issue #75, opened 2026-09-08
Rewrites discard unmodelled data: claudeAiOauth subkeys and unknown profiles.toml keys


ZviBaratz:
`serialize_credentials_preserving_extra` already states the correct rule, and its doc
comment is the best summary of the bug:

> Everything already in the file is kept … this only rewrites a store over itself, where an
> unrecognised key belongs to the account whose file it already is. **Dropping it here
> would lose data no other writer holds.**

0.15.x applies that rule to top-level blocks. It is not applied one level down, nor to
`profiles.toml`.

### 1. `claudeAiOauth` subkeys

```rust
// src/profile.rs — preserve_extra_blocks
for (key, extra) in existing_obj {
    if key == "claudeAiOauth" { continue; }
    value_obj.insert(key.clone(), extra.clone());
}
```

Siblings survive; everything inside `claudeAiOauth` is rebuilt from `OAuthToken`, which
models five fields and has no catch-all:

```rust
pub(crate) struct OAuthToken {
    pub(crate) access_token: String,
    pub(crate) refresh_token: Option<String>,
    pub(crate) expires_at: Option<i64>,
    pub(crate) scopes: Option<Vec<String>>,
    pub(crate) subscription_type: Option<String>,
}
```

Claude Code writes three more — `rateLimitTier`, `refreshTokenExpiresAt`, `clientId` — and
every clauth write of the store deletes them. `grep -rn 'rateLimitTier' src/` returns
nothing.

**Repro — deterministic, no network, no real account.** It does not take a token rotation;
any profile save is enough, and `clauth disable` is a pure config mutation:

```bash
F=$(mktemp -d); mkdir -p "$F/.clauth/profiles/fixture" "$F/.clauth/profiles/holder"
cat > "$F/.clauth/profiles.toml" <<'TOML'
active_profile = "holder"
profiles = ["holder", "fixture"]
fallback_chain = []
TOML
printf 'fallback_threshold = 95\n' | tee "$F/.clauth/profiles/fixture/config.toml" \
                                        "$F/.clauth/profiles/holder/config.toml" >/dev/null
cat > "$F/.clauth/profiles/fixture/credentials.json" <<'JSON'
{ "claudeAiOauth": {
    "accessToken": "PLACEHOLDER", "refreshToken": "PLACEHOLDER",
    "expiresAt": 1788999999000, "refreshTokenExpiresAt": 1790000000000,
    "scopes": ["user:inference"], "subscriptionType": "team",
[...]

PR #80, opened 2026-09-16
fix(login): stamp rateLimitTier from the /profile probe so plan-gated flags see the seat

### LINKED ISSUE #78: Team plan: sessions on a clauth-minted login hit Claude Code's "Fable 5.1 now uses usage credits" gate; a captured `claude auth login` does not
## What happened

After adding two Team-plan profiles with `clauth login`, every fresh interactive Claude Code session on the machine (both `clauth start <profile>` and a bare `claude` after `clauth <profile>`) refused Fable 5.1 with the dialog "Fable 5.1 now uses usage credits. You don't have usage credits yet." and fell back to Sonnet 5. The same login served Fable fine via `clauth start <profile> -p "..." --model claude-fable-5-1`, and a coworker in the same org was unaffected, so the account entitlement was intact.

Deleting both profiles, running a genuine `claude auth login`, and capturing that login into clauth fixed it immediately. The setup-token sidecar and `rolling-token` modes both showed the same gate.

Observed difference in the account profile Claude Code fetched under the clauth-minted login vs the genuine one: `hasExtraUsageEnabled` false vs true, and `organizationRole` / `workspaceRole` missing. Claude Code's interactive client gates Fable client-side on that kind of entitlement read (anthropics/claude-code#93840, #91488, #79360), and `-p` mode skips the gate, which explains the asymmetry.

## What you expected

A login minted by `clauth login` should be indistinguishable from `claude auth login` to Claude Code. `oauth_login.rs` requests the same 6-scope set and client id, so the difference is somewhere after the mint. Candidates I could not verify: `oauth.rs` refresh falling back to `REFRESH_SCOPES_FALLBACK` (no `org:create_api_key`) when a profile stores no scopes; `keychain.rs` `Keep` allowlist dropping account-scoped siblings on a switch; or `subscriptionType` stamping.

## Steps to reproduce

1. Team plan account (Premium seat, Fable included in plan). macOS.
2. `clauth login <profile>`; `clauth <profile>`.
3. Open `claude`, `/model`, pick Fable 5.1, send one message: credits dialog.
4. Same profile, `clauth start <profile> -p "say ok" --model claude-fable-5-1`: succeeds.
5. `clauth delete --yes <profile>`; `claude auth login`; mirror Keychain to `~/.claude/.credentials.json`; `clauth capture <profile>`; `clauth <profile>`; repeat step 3: Fable answers.

## clauth version

clauth 0.15.2

## Surface

CLI (`clauth login`, `clauth start`, switch), macOS Keychain

## OS

macOS 26.6.2 (25G83), arm64. Claude Code 2.1.258.

## Logs / output

Dialog text: "Fable 5.1 now uses usage credits / Fable 5.1 runs on usage credits, purchased separately from your plan. / You don't have usage credits yet. / Switch to Sonnet 5 and continue / Request usage credits from your admin"

## Notes

Happy to run a diff of the credential JSON shapes (keys only, no values) between a `clauth login` mint and a `claude auth login` mint if that helps narrow it.

## AI

Investigation and this write-up were done with Claude Code (Fable 5.1).

roErlend:
> **operator asked:** "The author doesn't have a team plan and asked if I could debug it and create a PR" for #78, then "create a PR without posting it, so I can preview it first". The ask evolved from a research pass into this branch.

Stamp `claudeAiOauth.rateLimitTier` on a `clauth login` mint and carry it into the rolling sidecar, so Claude Code evaluates its plan-gated feature flags for the real seat instead of an untiered account. Closes #78.

Claude Code writes `rateLimitTier` into the login block itself at login time, from `/profile` (`organization.rate_limit_tier`), and reads it back at startup as a GrowthBook targeting attribute alongside `subscriptionType`. The flag that switches Fable into credits-only mode (`tengu_saffron_lattice`) is targeted off for Team Premium (`default_claude_max_5x`). A clauth mint carried no `rateLimitTier`, so the flag evaluated to its default, on, and every interactive session opened with the "Fable 5.1 now uses usage credits" dialog while `-p` runs (which skip that dialog) were fine. Scopes turned out identical between the two logins, and the org's extra-usage setting was not the difference either; the root-cause write-up with the reproduction table is on the issue.

What changed:

- `profile.rs`: `RATE_LIMIT_TIER_KEY` plus `OAuthToken::rate_limit_tier()` / `set_rate_limit_tier()` over the `extra` catch-all from #75. The field stays Claude Code's, not part of the modelled struct.
- `usage/fetch.rs`: `LoginProfile` carries the org's raw `rate_limit_tier` (blank reads `None`), from the same `/profile` body the login already fetches. No new request.
- `oauth_login.rs`: `finish_login` stamps it next to `subscription_type`. The `credentials_from_token` comment used to say Claude Code adds `rateLimitTier` on its first save; it does, but the gate runs at startup before any save, so I reworded it.
- `claude.rs`: `rolling_projection` carries the chain's tier into the sidecar (and only that key), since a `clauth start` session's Claude Code reads the sidecar at startup too. The setup-token sidecar is out of reach: an inference-only mint cannot call `/profile` (that is anthropics/claude-code#79360).

Not done: backfilling `rateLimitTier` into already-minted profiles during the usage poll, which fetches `/profile` anyway. Existing users need one `clauth login <name>` to pick the stamp up. Happy to add the backfill here or in a follow-up if you want it.

No user-visible wording, key bindings, or defaults changed.

## agent

- Claude Code (claude-fable-5-1), operated by @roErlend
- ran: `cargo fmt --all -- --check` clean; `cargo clippy --locked --all-targets --all-features -- -D warnings` clean; `cargo test --locked --all-features` 3872 passed, 9 failed, every failure in `actions::tests::codex_*` / `delete_codex` / `switch_codex`; the same set fails on the untouched base `fdfee38` in this environment (macOS, no codex login), so pre-existing and unrelated to this change. Targeted: the 3 new tests plus the neighbouring login-probe, rolling-token, and extras tests pass (21/21). Reproduction I ran on the operator's machine against their Team Premium seat, operator reviewed the output: same access token via `CLAUDE_CODE_OAUTH_TOKEN`, `tengu_saffron_lattice` evaluates `{"enabled":false}` with `CLAUDE_CODE_RATE_LIMIT_TIER=default_claude_max_5x` and `{"enabled":true,"planLimitsEndDate":"2026-07-20T07:00:00Z"}` without. End to end on the patched release binary, isolated under a temp `HOME`/`USER` so the operator's real profiles and Keychain item stayed out of it: `clauth login e2e` (operator completed the browser login on a Team Premium seat) stored `subscriptionType: team` plus `rateLimitTier: default_claude_max_5x`; `clauth start e2e -p ... --model claude-sonnet-5` then cached `tengu_saffron_lattice` as `{"enabled":false}`. The interactive Fable dialog itself was not exercised on the patched binary; the flag value is the same one the genuine login produces and the unpatched mint did not.
- unsure: whether `finish_login` is the only mint path worth stamping. I left `write_session_token` (setup-token mint) alone because its scope cannot read `/profile`, and did not touch the usage-poll backfill. The `rateLimitTier` key name is taken from Claude Code 2.1.258's persisted credential; I did not check older builds.


uwuclxdy:
Thanks for the PR. The fix lands in the right place: the stamp rides the login's existing `/profile` probe and the #75 catch-all carries it, so no new request and no drift of the modelled fields.

Findings from the review:

- The `finish_login` stamp has no test, and the full-login door tests already serve a `/profile` fixture and assert the outcome, so the pin is nearly free. I'll add it maintainer-side along with the comment fixes below.
- Three comment nits: the `credentials_from_token` comment lists `clientId` as written on a first save, but the measured save shape wrote `rateLimitTier` + `refreshTokenExpiresAt` and not `clientId` (inline below); the `rolling_token` field doc doesn't name the tier among the carried fields; `login_profile_from_raw` says "no normalization beyond a blank check" while it trims.
- Existing profiles stay untiered until a `clauth login <name>`: accepted as a follow-up, not required here. The backfill is a new scheduler leg writing into the credential store under the state flock, which deserves its own round.
- Setup-token sidecars stay untiered: accepted as a documented limitation (an inference-only mint cannot call `/profile`). I'll record it in the docs at merge.

One domain note: the repo's notes say a missing `rateLimitTier` costs "a silent extra round trip plus a wrong branch, never a visible error". The #78 repro shows a visible gate, the Fable credits dialog, so the generalization needs narrowing to the consumers it was traced for. I'll update that at merge with your measurement.

Verified here: trial-merged into current `mommy`, the gate is green (3927 passed, 1 skipped, fmt/clippy/windows cross-lint/deny/audit), the three new tests pass, and breaking the rolling carry reds its test.

src/oauth_login.rs
```diff
@@ -635,6 +641,9 @@ fn finish_login(
         && let Ok(probe) = crate::usage::probe_login_profile(&oauth.access_token)
     {
         oauth.subscription_type = probe.subscription_type;
+        if let Some(tier) = probe.rate_limit_tier {
+            oauth.set_rate_limit_tier(tier);
```
uwuclxdy, line 679: minor: this stamp is the fix for #78 and no test asserts it, so the wiring can drop without a red. the paste/loopback door tests already serve a `/profile` fixture and assert the outcome, so the pin is one fixture field plus one assert, landing maintainer-side.

src/oauth_login.rs
```diff
@@ -512,9 +512,12 @@ fn credentials_from_token(token: crate::oauth::TokenResponse) -> ClaudeCredentia
             expires_at: Some((now_ms() + token.expires_in * 1000) as i64),
             scopes,
             subscription_type: None,
-            // A login clauth mints itself has no outside-written keys to keep;
-            // Claude Code adds its own (`rateLimitTier`, `clientId`) on its
-            // first token save, and the catch-all holds them from then on.
+            // A login clauth mints itself has no outside-written keys to keep.
+            // Claude Code adds its own (`clientId`, `refreshTokenExpiresAt`) on
```
uwuclxdy: nit: the measured save shape wrote `rateLimitTier` + `refreshTokenExpiresAt` and not `clientId`, which this enumeration now claims. consider dropping the enumeration.

src/oauth_login.rs
uwuclxdy: landed in `8723c7c0` on branch `pr80-landing` in this repo, along with the comment fixes; merging it into your branch is all that is left, I have no push access to the fork.

src/oauth_login.rs
uwuclxdy: landed with the other comment fixes in `8723c7c0` on branch `pr80-landing` in this repo.

src/oauth_login.rs
uwuclxdy: the branch pull landed — the pin rides the PR head now, and the threads below are resolved.

uwuclxdy:
Merged as `ef01edfb` (`fix(login): stamp rateLimitTier so plan-gated flags see the seat`), squash.

Merge-time notes: a `claude setup-token` mint reads as untiered to Claude Code, existing profiles pick the stamp up with one `clauth login <name>`, and the usage-poll backfill stays open as the follow-up.

